Issue #154, opened 2026-04-20
Try upgrading Godot to address Windows R2 TLS handshake failure


kion-dgl:
## Problem
Windows builds running Godot 4.5.1 fail every TLS handshake against Cloudflare R2 with `MBEDTLS_ERR_SSL_FATAL_ALERT_MESSAGE` (-30592):

```
ERROR: TLS handshake error: -30592
   at: _do_handshake (modules/mbedtls/stream_peer_mbedtls.cpp:88)
WARNING: [bootstrap] http 0 on https://pub-...r2.dev/packs/...
```

Every pack falls back to Arweave, so bootstrap completes — but each cold launch eats a multi-second handshake timeout per pack. Linux/Android are unaffected.

## What we already did
PR #153 (0.21.5) added a host blacklist in `scripts/2d/bootstrap.gd`: when R2 fails the TLS handshake on the first pack, subsequent packs skip R2 entirely and go straight to Arweave. This trims the per-launch overhead from ~5x failed handshakes to 1x.

That's a workaround, not a fix. R2 is still unreachable from Windows builds.

## Hypothesis
Godot 4.5.1's bundled mbedTLS can't negotiate with Cloudflare's TLS 1.3-preferred edge. `TLSOptions.client_unsafe()` is already set to skip cert validation (`bootstrap.gd:54`) — that addresses `-9984` cert errors but not protocol-level handshake aborts. Newer Godot releases have picked up mbedTLS updates and TLS bugfixes.

## Spike plan
- [ ] Check the latest Godot 4.x stable release (likely 4.6+) and read its changelog for mbedTLS / HTTPS / TLS-related entries
- [ ] Test a Windows export against R2 with the newer engine (no other code changes) and confirm whether the handshake succeeds
- [ ] If it works: plan the 4.5.1 → 4.x upgrade — engine bumps can ripple into shader compilation, import format, GDScript parser changes, so do it on a branch and run the full test suite + a manual play-through
- [ ] If it doesn't work: dig into the mbedTLS handshake details (what cipher / protocol version is it offering vs. what Cloudflare expects), consider building Godot with a system mbedTLS, or evaluate alternative HTTP backends

## Acceptance
- Windows bootstrap log shows `[bootstrap] mounted user://packs/...` for every pack via R2 (no `TLS handshake error: -30592`, no Arweave fallback for the common case)
- Linux/Android/web behavior unchanged

## Context
[...]

PR #186, opened 2026-04-24
0.25.0 — Godot 4.6.2 upgrade + invert-camera-X default flip

kion-dgl:
## Summary
- Bumps engine from Godot **4.5.1 → 4.6.2** across CI/release workflows, template cache keys, editor-settings paths, and `project.godot` config/features. Addresses #154 (Windows R2 TLS handshake failure suspected to be engine-side).
- Flips **Invert Camera X** default from ON to OFF and inverts the operator in `orbit_camera.gd` so the natural orbit direction is the OFF behavior. Matches playtest feedback (Rozalin) that toggle defaults should default OFF; players who want flight-stick-style inversion can enable it explicitly.
- Bumps `VERSION` and `project.godot` config/version to **0.25.0** (minor — engine bump is notable).

## Test plan
- [ ] CI: lint + headless validation pass on 4.6.2
- [ ] CI: web export succeeds on 4.6.2
- [ ] CI: Android export succeeds on 4.6.2 (templates re-cached under new key)
- [ ] CI: release workflow dry-pass on 4.6.2
- [ ] Local: `godot --headless --quit` boots with no script errors (verified — boots as `0.25.0 on Linux/arm64 (Godot 4.6.2-stable)`)
- [ ] Manual: launch in editor, walk into a stage, confirm camera rotates the expected direction with Invert Camera X off
- [ ] Manual: toggle Invert Camera X on, confirm direction reverses
- [ ] Existing playtester saves: `invert_camera_x=true` will now produce the inverted behavior. Players with old saves need to toggle once. Acceptable for pre-alpha.

## Out of scope (followups from same Rozalin batch)
- "Debug menu off by default" — needs design decision (hide debug toggles from Options menu behind a separate Show Debug Options toggle?). Not addressed here.
- Right-stick start-menu scroll not working on DS4 — code in `pso_start_menu.gd` looks correct (axis 3 / hysteresis); likely DS4-specific axis mapping needs a device probe. Not addressed here.


kion-dgl:
Closing without merge. Goals weren't met:

- Godot 4.6.2 upgrade did **not** fix #154 (mbedtls TLS handshake against Cloudflare still fails identically — error is upstream Godot vs Cloudflare's edge TLS profile, engine bump didn't change it).
- 4.6.2 also introduces a T-pose animation regression with the current asset pack (GLB/animation compat issue) — would need investigation before any future upgrade attempt.
- Worker / DigitalOcean proxy experiments confirmed the issue is not host-specific and isn't worked around by any Cloudflare-fronted route.

Useful side effects from this work that we'll fold in separately:
- Invert Camera X default flip (Rozalin feedback) — file as its own small PR.
- Publish pipeline now Arweave-only (R2 was dead weight for clients given mbedtls). Will be its own follow-up.
- R2 bucket cleaned (old packs purged); R2 still used for per-file dev asset sync only.

Branch left on origin for reference; can be deleted.

PR #200, opened 2026-05-03
0.27.3 — Storage, equipment, and unarmed-combat polish

kion-dgl:
Five small fixes shaken out of testing v0.27.2 on the Pixel 8a + Retroid Pocket 3+. All verified working on a release-signed local build before commit.

## Summary

| # | What | File(s) |
|---|---|---|
| 1 | "Exit to City" trigger moved from `(0.04, 1, +3)` to `(0, 1, -3)` | `scripts/3d/city/city_underground_controller.gd` |
| 2 | Storage D-pad Left/Right swaps panels (not tabs); palette_swap (LB) toggles tabs | `scripts/2d/storage.gd` |
| 3 | Equipped weapon/armor/mag locked from being moved to storage; greyed out in UI with "Unequip first!" hint | `scripts/2d/storage.gd` |
| 4 | Unarmed combat now uses `common_m/w.glb` + `pmbn/pwbn` (barehanded) animations instead of falling back to saber | `scripts/3d/player/player.gd` |
| 5 | Mobile on-screen dpad now shows on the first-launch controller picker | `scripts/autoloads/mobile_controls.gd` |

### 1. "Exit to City" position
Previous location was at the wrong end of the underground room near Photon Collector / Blackjack — players had to walk into the wall to find it. Trigger volume centred at `(0, 0, -3)` on the floor; Y=1 raises the box to torso height so the prompt fires reliably.

### 2. Storage tab nav
`ui_left`/`ui_right` blindly toggled between Items and Meseta tabs, even when the player was in the middle of moving items between inventory and storage panels. Once items landed in storage there was no obvious way to focus the storage panel and pull them back. Now matches the visual layout: panels switch with Left/Right, tabs switch with palette_swap (LB / Shift) — the same convention used in fieldMenu per `web/src/controls/ControlsDiagram.tsx`. Meseta tab also accepts Left/Right as a way back to Items so users who don't know about LB aren't stuck.

### 3. Storage equipped-item lock
Same rule the shop's sell flow already enforces (`scripts/2d/shops/item_shop.gd:_sell_selected`): an equipped item can't leave the inventory without unequipping first. Initially I auto-unequipped on transfer, but you pushed back: silently mutating equipment from a storage screen is surprising. Lock-with-hint matches the existing UX. Equipped items render in `PszStyle.TEXT_MUTED` with the existing `[E]` tag so the locked state is legible at a glance.

### 4. Unarmed barehanded animations
[...]

PR #201, opened 2026-05-03
0.27.4 — Cert-pin hard-fail + three-way version sync

kion-dgl:
Two infrastructure followups deferred from PR #198, now feasible:

## 1. Hard-fail apksigner cert pin (`release.yml`)

The previous "log only" cert dump is now a strict comparison against the pinned PSZ release cert SHA-256:
`2b95171ddb22ca5e6e28ae4ad5003e824e6c438cd6e38f0a253f7c0144a4fb06`

Parses apksigner's `SHA-256 digest:` line (format confirmed against the v0.27.2 release run log: `V3.0 Signer: certificate SHA-256 digest: 2b95171d...`), takes the first match across signature scheme versions (V1.0/V2.0/V3.0/V3.1 — all signed by the same key in our setup), lowercases + strips, compares.

Fail modes:
- **Mismatch** — wrong `ANDROID_KEYSTORE_B64` secret in the prod environment, or someone rotated the keystore without updating the pin. Error message points at both possibilities.
- **Parse failure** — apksigner output format change, or the APK is unsigned. Either way, fail-fast is better than silently shipping unverifiable bytes.

## 2. Three-way version sync enforcement (`ci.yml`)

The `version/name="0.2.2"` bug in `export_presets.cfg` was the smoking gun: hardcoded to "0.2.2" forever, so Android `dumpsys package`, Play Console, and any future code-signing audit all under-reported the version regardless of what `VERSION` said. We caught it when the phone reported "0.2.2" while running 0.27.x.

**Fixed the value** to 0.27.4 in this PR. **To prevent the same drift in future**, the `version-check` CI job now also asserts:
- `project.godot config/version` == `VERSION`
- `export_presets.cfg version/name` == `VERSION`

Bumping `VERSION` without bumping the other two now hard-fails the PR with a specific pointer to which file is out of sync. Same forcing-function pattern as `verify-assets` and the existing version-bump check.

## Sample CI output

```
Version bumped: 0.27.2 → 0.27.4 ✓
  VERSION                          = 0.27.4
  project.godot config/version     = 0.27.4
  export_presets.cfg version/name  = 0.27.4
```

Bad case (sample):
```
::error::export_presets.cfg version/name='0.27.3' must equal VERSION='0.27.4' (Android dumpsys reads this)
```

## Versioning

Bumped `0.27.2 → 0.27.4`, **skipping 0.27.3** which is owned by PR #200. This way both PRs can land in either order without a `VERSION` conflict — whichever lands first becomes main, the other rebases cleanly.

## Test plan

- [x] CI `version-check` job passes — verifies all three are in sync at 0.27.4 (visible in this PR's CI run)
- [ ] After merge, `release` job runs the new cert-pin check and prints `✓ APK signed with pinned release cert (SHA-256 ...)` 
- [ ] Manual: try a PR that bumps only VERSION but not the other two → should hard-fail with the new error message (not part of this PR; verify next time someone forgets)

## Out of scope

- `version/code` (Android integer build number) is still hardcoded to `1`. For sideload installs that's fine (`adb install -r` accepts equal codes), but if PSZ ever ships to Play Store this needs auto-bumping too. Tracked mentally; leaving for later.


kion-dgl:
Re-ported onto current main as #206. The two infra changes (cert-pin hard-fail + three-way version sync) are preserved verbatim; the stale VERSION bump (0.27.4) and accidental asset deletions are dropped, and the 'Mark exports as CI builds' step that landed in #202 after this branch was cut is preserved. Closing in favor of #206.

PR #211, opened 2026-05-09
TEST: swap item-shop NPC for VRM model

kion-dgl:
## Summary

Replaces the item-shop NPC (was `np_003_00_0.glb`) with a VRM model (renamed to `.glb` so Godot's GLTF importer picks it up). Living under `assets/ui/item_shop_vrm/` so it ships in the APK directly without needing a pack republish.

## Why

Quick spike to see how a VRM-style anime model reads against PSO-style NPCs. If it looks plausible we can pull the rest of the shop NPC art in this direction; if not we keep it as a one-off curiosity.

## Caveats

- Idle animation cleared — `pso_f_sh_stand` targets PSO bones the VRM rig doesn't have. Model stands in its bind pose. If we keep the swap we'll need to retarget or use a VRM-compatible idle.
- Adds ~12 MB to the APK from the GLB + 25 textures Godot extracted. Acceptable for a test; for permanent use we'd move it back into the asset pack.
- VRM-specific features (expressions, spring bones, first-person look) silently dropped by Godot's GLTF importer — it only sees the mesh + base skeleton.

## Test plan

- [ ] Build APK, install on Retroid, walk into the market, eyeball the item-shop counter.
- [ ] Confirm the existing weapon-shop and grind-shop NPCs are unchanged.
- [ ] Confirm shop interaction still triggers (the `target_scene_path` wiring is independent of the model).


kion-dgl:
Closing — POC validated the VRM-as-GLB import path works, but the model stands in T-pose without the PSO idle anim, so this isn't ready to merge. Tracking the actual retargeting work in #212.

PR #246, opened 2026-05-30
Fix dev asset fetch on macOS (BSD xargs command-size limit)

kion-dgl:
## Summary

`scripts/tools/fetch_assets_dev.sh` **silently downloads almost none of the assets on macOS**, then exits as if it succeeded — leaving a half-populated `assets/` and a project that won't run.

Root cause: BSD `xargs` (the macOS default) enforces a small per-invocation command-line size limit when using `-I {}` substitution. The long inline `bash -c` body overflows that limit, causing `xargs` to abort mid-stream. This never surfaced in CI because Linux runners use GNU `xargs`, which has no such limit.

## Changes

- **`scripts/tools/fetch_assets_dev.sh`**: Switch parallel dispatch from `-I {}` to NUL-delimited `xargs -0 -n 1`, passing each manifest line as `$1` with no template substitution. Use `${1-}` instead of `$1` in the wrapper to avoid an unbound-variable error under `set -u` when `xargs` fires on an empty manifest. Added an empty-record guard in `download_one`.
- **`BUILD.md`**: Document that the script runs unmodified on macOS and Linux; scope the "no extra tooling" claim to `xargs`/`findutils` (no `gxargs` needed); keep the `md5sum` (`brew install coreutils`) requirement explicit.
- **`VERSION`**, **`project.godot`**, **`export_presets.cfg`**: Bump version `0.30.19` → `0.30.20` across all three sources required by the CI `version-check`.

## Testing

- [x] Player movement works
- [x] Map transitions work
- [x] No console errors
- [x] Feature-specific tests:
  - [x] Full macOS run from clean `assets/`: all 21342 manifest files downloaded, 0 failures (~500 MB)
  - [x] Dispatch tested in isolation with lines containing spaces and tabs
  - [x] Confirmed `\n`→`\0` conversion produces no phantom empty record
  - [x] CI `version-check` passes with all three version sources synced to `0.30.20`


kion-dgl:
Thanks @copilot — addressed all three:

1. **`set -u` unbound `$1`** (`fetch_assets_dev.sh`): switched the wrapper to `download_one "${1-}"` / `echo "${1-}"` (commit b38e157e). On an empty manifest GNU xargs fires once with no arg; `${1-}` keeps both the GNU and BSD paths safe and lets `download_one`'s empty-line guard handle it.
2. **BUILD.md contradiction**: reworded so the "no extra tooling" claim is scoped to `xargs`/`findutils` (no `gxargs`), with the `md5sum` (coreutils) requirement kept explicit (commit b38e157e).
3. **`export_presets.cfg` version/name**: already bumped to `0.30.20` in commit 72772cf3 (your review ran against the first commit). `version-check` is green now.

PR #252, opened 2026-06-03
quest graph canonicalization + finding_ogi section A unblocks + autopilot HUD

kion-dgl:
## Summary

Three semi-independent pieces of work, all aligned with the [story-progression spec](http://192.168.10.36:4000/states/story-progression):

### 1. Canonicalize the quest graph

Nine quests had missing or wrong `parent_quest` fields per the spec graph; corrected. Manifest reordered into a valid topological sort so each quest follows its parent — important for the autopilot's "scroll N times" guild-menu navigation and matches the order a player actually unlocks quests. (`5f0e8b45`)

`SR PP AS DOE CAS NR PW FO SIS MFP IT HR CSY TBS SMM DC` + variants tail.

Side effect: previously `static_in_the_snow` sat at index #4 with empty `parent_quest`, so it could be autopilot-tested in isolation. With the corrected graph it's behind FO — which is still being unblocked (see below).

### 2. Unblock finding_ogi section A end-to-end

Three issues, all data-or-runtime, none autopilot-side:

- **Wave 1 data fix** (`4ae02cfe`) — cell B 1,2 (s03a_na1) had its two enemies on `wave: 3` with empty waves 1 and 2. Wave 1 spawned with 0 enemies, the wave system silently stalled (no `died` signals to advance), wave-3 enemies never spawned, the key drop hook never fired. Real-player-breaking. Moved both enemies to wave 1.
- **Empty-wave runtime guard** (`4ae02cfe`) — when a wave spawns zero enemies, `_spawn_wave` now calls `_check_room_clear()` immediately to auto-advance. Prevents any future cell with the same shape from silently breaking.
- **Section A exit_direction fix** (`4ae02cfe`) — section A said `exit_direction: "east"` but end cell 0,0 had `warp_edge: "west"`. Every other passing quest has these match. The mismatch made the autopilot exit east from 0,0 and ping-pong forever between 0,0 and 0,1 (its only eastern neighbor). Changed `exit_direction` to "west".
- **Scheduler self-drop preference** (`03d81229`) — cell B 3,2 in finding_ogi is a hub gated by three doors with four cells dropping keys for it (self + three spokes). The scheduler's `keyDropOf` map was last-write-wins, so it concluded "the key for B 3,2 is at B 1,4" — past the locked gate, circular dependency. Switching to first-write-wins makes the self-drop win and unblocks the first gate.

With these in place, finding_ogi section A clears, transitions through E, reaches section B start, opens the first gate at B 3,2. Section B still gets stuck at the second locked gate — that needs a real planner upgrade (see follow-up PR for the solver refactor).

### 3. HUD debug overlay under PSZ_AUTOPILOT

`scripts/3d/field/field_hud.gd` had a `_DebugInfoPanel` class but it was never instantiated and `set_debug_info()` was a no-op. Wired it up gated on `PSZ_AUTOPILOT=1`. Shows `MM:SS | quest_id | Section | Cell pos` at top-center. Makes it trivial to cross-reference a video timestamp against the sanity log. (`dcf2a9c4`)

## Version

`0.31.2` → `0.31.3` (patch).

## Verified

- finding_ogi run #4 (post-self-drop fix): section A complete (13/13 cells + transition), section B start (s03b_sa1) reached, B 3,2's first gate opened successfully on visit 1. Stuck at the *second* gate awaiting the multi-gate hub planner.
- `bun scripts/tools/quest_plan.ts --all` clean — no scheduler warnings.
- Other quests' plans unchanged by the self-drop preference.


kion-dgl:
Fixed in 4263b9b1. The duplicate `_process` was indeed silently overwriting the FPS counter / overlay-visibility behaviour — Godot doesn't error on duplicate methods at the class level, just takes the last definition. Merged the autopilot HUD overlay's wall-clock render into the original `_process` so both behaviours coexist.

Issue #294, opened 2026-06-08
Dedup near-identical GDScript functions (codebase-graph SIMILAR_TO clusters)


kion-dgl:
The codebase-memory graph's `SIMILAR_TO` edges (jaccard ≥ 0.85) flag near-identical GDScript functions — the same signal that surfaced the #215 dup families. These are extractable into shared helpers/bases. Concrete clusters (query: `MATCH (a:Function)-[r:SIMILAR_TO]->(b:Function) WHERE r.jaccard >= 0.85 ...`):

**Element setup family** (biggest win — a shared `Element` base or mixin):
- [x] `_setup_wall_collision` / `_setup_box_collision` / `_setup_enemy_collision` / `_setup_fence_collision` — `wall.gd` / `box.gd` / `enemy_spawn.gd` / `fence.gd` (j=1.0) — **done in #312** (inc 1, GameElement)
- [x] `_setup_textures` — `wall.gd` / `box.gd` (0.95) — **done in #334** (GameElement._setup_mirror_textures)
- [x] `_setup_laser_material` / `_setup_scroll_material` — `gate.gd` / `key_gate.gd` / `message_pack.gd` (0.95–1.0) — **done in #327** (inc 6)
- [x] `_setup_prompt` — `message_pack.gd` / `key_gate.gd` / `key_pickup.gd` (1.0) — **done in #326** (inc 5)

**Input handlers** (shared list-nav base):
- [x] `_unhandled_input` — `mag_list.gd` / `mag_feeder.gd` / `inventory_screen.gd` (1.0) — **done in #332** (ShopNav.handle, #274 inc 3)

**Texture apply** (shared util):
- [ ] `_apply_texture` — `weapon_test.gd` / `companion_npc.gd` / `enemy_base.gd` (1.0) — cluster has since grown: `field_npc.gd` / `enemy_spawn.gd` also carry copies
- [x] `_apply_texture_recursive` — `character_create.gd` / `character_select.gd` (1.0) — **done in #325** (inc 4, MeshUtils)

**Same-file near-dupes** (parameterize):
- [x] `_add_weapon_shop_cart` / `_add_item_shop_cart` — `city_market_controller.gd` (1.0) — **done in #315** (inc 2)
- [x] `_get_visible_slots` / `_get_visible_slot_names` — `equipment_screen.gd` (1.0) — **done in #315** (inc 2)
[...]

PR #312, opened 2026-06-09
refactor: dedup element static-collision setup into GameElement (#294)

kion-dgl:
## Why

First increment of #294 (SIMILAR_TO dedup) — the **biggest cluster**. `wall.gd`/`box.gd`/`fence.gd`/`enemy_spawn.gd` each had a byte-identical `_setup_*_collision` that built a `StaticBody3D` environment collider (layer 1, mask 0, box sized to `collision_size`), differing **only by the node name** (j=1.0, 6 near-dup pairs).

## What

Folds them into a shared `GameElement._build_static_collision(name) -> StaticBody3D` that builds + attaches the body and returns it. Each subclass now does:
```gdscript
collision_body = _build_static_collision("WallCollision")
```

**Why this is safe re: the #283/#284 Android-export base-class bug:** that bug was about introducing a *new* cross-script base for the standalone shops. Here the helper goes on the **existing** `GameElement` base that every element already `extends` (and which the autopilot matrix exercises on all platforms). `collision_body` stays a per-subclass `var` (the helper *returns* the body), so there's no base-member migration and `gate`/`key_gate` are untouched.

Node names are preserved (`WallCollision`/`BoxCollision`/`FenceCollision`/`EnemyCollision`) — verified none are referenced via `get_node` anywhere.

## Verification

- Dup baseline **162 → 156** pairs (the collision family gone); dead-count unchanged.
- Headless suite: **1462 passed, 0 failed**.
- ⏳ Full autopilot matrix pending before merge (touches field-element collision — the matrix is the real guard).

Follow-up #294 clusters (separate increments): `_setup_textures`/`_setup_*_material`/`_setup_prompt`, shared list-nav `_unhandled_input`, `_apply_texture(_recursive)`, same-file parameterize pairs, `_item_fits_slot`, `point_in_triangle`.


kion-dgl:
🤖 **Autopilot regression matrix running** (local dev-box gate — Copilot reviews are off, so the full matrix is our pre-merge gate). Running the full chain: Boot → SR → PP → AS → DOE → FO → SIS → IT → heretic → control_system → the_broken_seal → dark_castle → shops. Plus the new `test_element_collision_setup` is in the headless CI suite. Will post results here when it completes (~40 min).

kion-dgl:
⚠️ **Matrix hit a failure — investigating (likely a flake, not this PR).**

`as_canon` failed with a **stuck-walk in stage `s02a_ga1`** (cell 0,2, dist=3.61, "author waypoints for this stage"). Notably the floor check passes (3/3 hits) and the diagnostic is a **navigation/waypoint** stall — the same signature as the heretic/csy waypoint issues, *not* a collider failure.

Why I believe this is **not** caused by this PR:
- The dedup is behaviorally identical — `_build_static_collision` produces the same `StaticBody3D` (name, layer 1, mask 0, box of `collision_size`, same y-offset), created at the same point in `_ready`. It changes no collider geometry.
- A stuck-walk needs a blocking collider (unchanged) or a waypoint gap (untouched) — neither is affected here.
- `as_canon` passed in every prior matrix run (batches 1, 2, and the #307+#308 run); its siblings `as_moon`/`as_sol` passed in *this* run.

Re-running `--from as` to confirm it clears. Will post the re-run result before any merge.

kion-dgl:
## Matrix results — PR is sound, but the matrix flaked (not this PR)

**Run 1:** all green except `as_canon` (stuck-walk `s02a_ga1`, dist=3.61).
**Run 2 (`--from as`):** `as_canon` ✅ cleared, `as_moon`/`as_sol`/`doe`/`fo`/`sis`/`it` ✅ — but `heretic` flaked (stuck-walk `s06a_td1`, dist=3.65, the same west-exit approach #308 fixed and that passed in the #307+#308 and batch-2 runs).

Across both runs **every quest passes at least once**, and the failures move (as_canon → heretic) — a textbook intermittent flake, not a regression.

**Why it's not this PR:** both failures are navigation stalls (floor present 3/3, no collision/script error) ~3.6 units from an exit waypoint. The dedup is behaviorally identical — `_build_static_collision` builds the same `StaticBody3D`; it changes no collider geometry and can't cause a nav stall. The new `test_element_collision_setup` pins that in CI.

**Systemic flake spotted:** stuck distances this session are 3.61 / 3.63 / 3.63 / 3.65 — all the *final approach to an exit/spawn waypoint*. This intermittent ~3.6-unit stall is what makes the matrix red spuriously; it needs a harness retry and/or a walk-primitive fix (tracking separately). Holding merge pending maintainer call on the gate.

Issue #313, opened 2026-06-09
Autopilot robustness: intermittent ~3.6-unit exit-approach stuck-walk flakes the regression matrix


kion-dgl:
## Problem

With Copilot reviews gone, the autopilot regression matrix is our pre-merge gate ([[feedback_tests_are_the_merge_gate]]). But it's **intermittently flaky**: quests fail with a stuck-walk on the **final approach to an exit/spawn waypoint**, then pass on a clean re-run. This spuriously reds good PRs.

## Evidence (this session)

Stuck distances are suspiciously consistent — always ~3.6 units from the target exit/spawn waypoint:

| Quest | Stage | dist | Notes |
|---|---|---|---|
| heretic | s06a_td1 | 3.63 / 3.65 | failed, then passed after #308 + on re-run; failed again intermittently |
| control_system | s06b_lb3 | 3.63 | failed, fixed via #308-style waypoints, passes |
| apothecary_supply (as_canon) | s02a_ga1 | 3.61 | failed run 1, **passed clean re-run** (PR #312 gate) |

Common signature: `[sanity] stuck-walk ... dist≈3.6`, **floor present (3/3 hits)**, pure-axis direction toward an exit/spawn placed ~3–4 m inward from the gate, then `FAIL: walk stuck ... author waypoints`. It's **not** collision/script error and **not** stage-specific — it moves between quests run-to-run, and correlates with machine load (3× sim speed + concurrent godot instances).

## Two-pronged fix

**1. Harness retry (quick, makes the gate reliable) — pick this up first**
- `scripts/tools/autoplay/run_regression_matrix.sh`: on a stuck-walk failure, **retry the quest 1–2×** before declaring it failed. Only a *consistent* fail reds the gate; intermittent flakes self-heal.
- Log each retry clearly (so a flaky stage is still visible as "passed on retry N", not hidden).
- Keep it scoped: retry only on the stuck-walk/`rc=124`-style failure signature, not on real `[sanity] FAIL` asserts (e.g. a quest-logic mismatch should fail immediately).

**2. Walk-primitive root cause (deeper, separate increment)**
- The walk primitive jams ~3.6 m from exit/spawn waypoints on final approach. Investigate the arrival/advance logic in `autopilot.gd`'s walk-to-waypoint (arrival radius vs the inward-placed spawn target, or a near-gate collider). Fixing it helps the **real in-game** experience too, not just the harness.
- Add a characterization test where feasible.

## Why it matters
[...]

PR #314, opened 2026-06-09
ci(autopilot): retry intermittent stuck-walk/timeout flakes in the matrix (#313)

kion-dgl:
## Why

Closes part of #313. With Copilot reviews off, the autopilot regression matrix is the pre-merge gate — but it's intermittently flaky: quests stuck-walk on the final approach to an exit/spawn waypoint (~3.6 units out — **heretic, csy, and as_canon all hit it this session**, stuck distances 3.61/3.63/3.63/3.65), then pass on a clean re-run. That spuriously reds good PRs (it nearly blocked #312, which was sound).

## What

`run_godot` now **retries a failed quest** up to `QUEST_MAX_ATTEMPTS` (default 3) **only when the failure is a retryable flake**:
- a `[sanity] FAIL: walk stuck` navigation stall, or
- a timeout/hang (`rc=124`).

Each retry **re-stages the userdir from a clean seed copy** (a stuck run mutates the save, so a naive re-run would resume from bad state). **Real failures fail immediately** — logic `[sanity]` asserts, parse/script errors, other non-zero exits. The per-quest sidecar gains an `attempts` field, and a pass-after-retry is logged explicitly (flakes stay visible, not hidden).

## Tests — `--self-test` (CI-runnable, pack-free)

New `bash scripts/tools/autoplay/run_regression_matrix.sh --self-test` fixture-checks the `_is_retryable_flake` classifier:

```
PASS: stuck-walk (rc=1) → retry
PASS: timeout (rc=124) → retry
PASS: stuck-walk also wins under rc=124
PASS: real [sanity] FAIL assert (rc=1) → no retry
PASS: parse/script error (rc=1) → no retry
PASS: clean-ish non-stuck rc=1 → no retry
── self-test: 6 passed, 0 failed ──
```

`bash -n` syntax-clean.

## Scope note

This makes the **gate trustworthy**. The deeper **root-cause fix** for the ~3.6-unit walk-primitive exit-approach stall (which would help real in-game autopilot too) stays tracked in #313 as a separate increment.


kion-dgl:
🤖 **Autopilot matrix running** (local gate) — this run exercises the new retry logic itself. If one of the intermittent ~3.6-unit exit-approach stuck-walks fires, you'll see `flake on attempt N — retrying from clean save` followed by a pass (the fix working); otherwise it's green outright. Self-test already green (6/6). Will post results when complete (~40 min).

kion-dgl:
## ✅ Matrix results — fully green (16/16)

Boot, SR, PP (canon+backtrack), AS (canon+moon+sol), DOE, finding_ogi, SIS, IT, heretic, control_system, the_broken_seal, dark_castle, shops — **all pass**.

This run, no intermittent flake fired, so retries didn't need to engage — but the chain phases now log `attempt 1/3`, confirming the retry harness is live, and the `--self-test` proved the classifier (stuck-walk/timeout → retry; real asserts/parse errors → fail-fast, 6/6). Notably `as_canon` and `heretic` — which flaked during #312's gating — passed first attempt here.

Net: the gate now self-heals the ~3.6-unit exit-approach flakes (up to 3 attempts from a clean save) instead of spuriously redding good PRs, while still failing fast on real regressions. Merging.

PR #336, opened 2026-06-12
fix(combat): collapse element→status maps — poison/slow/paralysis now proc (#242)

kion-dgl:
## Why

Closes #242 — first PR of the combat window. `CombatManager` carried two divergent element→status maps: `ELEMENT_STATUS` (multi-status, dead) and `ELEMENT_TO_STATUS` (single-status, the wired one). Because the live map routed `dark→devil` / `light→sleep` and nothing else, **poison, slow, and paralysis were fully implemented but unreachable** through element procs.

## Canonical routing (now in spec /states/enemies)

| Element | Procs |
|---|---|
| fire | burn |
| ice | freeze *or* slow |
| lightning | stun *or* paralysis |
| dark | poison *or* devil (instant ¼ HP) |
| light | sleep |

One map (`ELEMENT_STATUS`), one roll function (`roll_element_status()` — uniform pick), both consumers (`EnemyBase._try_apply_status`, `CombatManager._try_element_special`) route through it. `ELEMENT_TO_STATUS` is deleted. Every `STATUS_EFFECTS` entry is reachable; devil stays dark's instant effect per the spec.

## Also fixed (found while here)

- `STATUS_COLORS` was missing the **paralysis** tint (the #242 secondary).
- `_try_element_special`'s proc message was keyed by *element*, so ice always announced "Freeze" even when slow procced — now names the rolled status.

## Test

`test_element_status` (seeded, per the #317 pattern): map invariants in both directions (every routed status defined-or-devil, every timed status reachable), tint coverage for all seven statuses, 200-sample rolls observing every option per element, and the live dark path landing **both** poison and devil with devil's ¼-HP verified.

## Verification

- Headless suite: **1580 passed, 0 failed** (35 new assertions)
- Ratchet: 0 new dup/dead/complexity/coupling; warnings gate clean
- ⏳ **Full autopilot regression matrix running now** (combat-window policy) — results posted before merge

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #337, opened 2026-06-12
feat(player): close state-machine gaps — i-frames, charge-cancel, SPRINTING removal (#273)

### LINKED ISSUE #273: Player state-machine gaps (from /states/player-state spec)
Surfaced while writing the `/states/player-state` behavior spec (PR #265). Three decided-but-unimplemented items in `scripts/3d/player/player.gd`'s state machine — these are gameplay *correctness*, distinct from the player.gd split tracked in #215-D. The spec page documents each as a "current gap" with the intended behavior.

- [ ] **Remove dead `SPRINTING` state** — `PlayerState.SPRINTING` has a speed/anim/footstep handling but **no code path ever transitions into it** (only demotes out). Drop it from the enum + `_handle_movement`. Locomotion is IDLE / WALKING / RUNNING.
- [ ] **Dodge i-frames not implemented** — `/mechanics/dodge` specifies an invincibility window during the roll, but `_handle_dodge` has no damage-immunity; `take_damage` can still knock a DODGING avatar into DAMAGED. Add the i-frame window (move phase).
- [ ] **Charge not cancelled on dodge** — a mid-charge technique is cancelled entering DAMAGED/DOWN but **not** when entering DODGING. `transition_to` should also release the charge on DODGING.
- [ ] (Keep `STUNNED` as a reserved/not-yet-implemented status state — documented, no action needed.)

Reference: https://github.com/kion-dgl/psz-godot/blob/main/spec/src/pages/states/player-state.astro (the "Planned changes (refactor punch-list)" section). Good to land alongside the `player.gd` split (#215-D) since they touch the same code.

🤖 Filed by Claude Code from the pre-beta behavior-spec pass.

kion-dgl:
## Why

Closes #273 — second combat-window PR. Three spec-vs-code gaps from `/states/player-state`, all documented there as intended behavior with "current gap" notes (now flipped to implemented).

## What

1. **SPRINTING removed.** The enum value had a speed, animation, and footstep handling, but no code path ever transitioned *into* it. Dropped from the enum, `_handle_movement`, footstep intervals, the animation table, and the sprint-anim resolution in `_load_weapon_animations`; `SPRINT_SPEED` / `FOOTSTEP_SPRINT_INTERVAL` deleted. Locomotion is exactly IDLE / WALKING / RUNNING. (companion_npc's `>= 1` is-moving check is ordinal-shift safe; comment updated. Nothing persists raw state ints.)
2. **Dodge i-frames** (`/mechanics/dodge`). `take_damage` is a no-op while DODGING during the **move phase** (`dodge_timer < dodge_move_end`); the recovery crouch + rise stays vulnerable, so the window is the roll itself, not the whole state.
3. **Charge-cancel on DODGING.** `transition_to`'s mid-charge release (clear slot, end charge visual, emit `tech_charge_released`) now also fires entering DODGING — previously only DAMAGED/DOWN, so rolling kept a phantom charge.

## Test

`test_player_states` pins all three:
- exact `PlayerState` table vs the spec (SPRINTING can't quietly return; STUNNED stays reserved), plus removed-const checks
- both sides of the i-frame window (move phase ignores a 25-damage hit; recovery takes a hit) via an off-tree player instance
- DODGING releases the charge with the right slot signal; a locomotion transition keeps it (control)

## Verification

- Headless suite: **1589 passed, 0 failed**
- Ratchet clean; warnings gate clean; spec pages updated (dodge status, invariants, punch-list)
- ⏳ **Full regression matrix running** (combat-window policy) — results posted before merge


kion-dgl:
## ✅ Full regression matrix: 16/16 phases pass

Run window 14:41–15:19Z (~39 min), report verified fresh. Boot through dark_castle plus all variants and shops — every phase green on the first attempt with the i-frame and charge-cancel changes live. Durations within a few seconds of the #336 baseline run, so the dodge changes didn't perturb autopilot combat timing.

Proceeding to the sanity merge gate.

Issue #318, opened 2026-06-10
Canon quest rewards unimplemented — report_quest must grant rewards (0/20 quests define them)


kion-dgl:
## Gap

The story-progression spec (`/states/story-progression`) states:

> **Report** — turning the quest in at the guild counter **MUST** persist it (`complete_mission` → `completed_missions`), **grant rewards**, and return to Free Roam.

But canon quest rewards are **not implemented**:

- **0 of 20** `data/quests/*.json` define a `rewards` field.
- `guild_counter.gd` emits `"rewards": {}` placeholders (lines 76, 94, 153).
- The only reward data in the repo lives in the **dead** `MissionRegistry` `.tres` (legacy early-CLI missions), whose ids match no canon quest — so it cannot be reused. See #281 (dead-code removal) for why that path is being deleted, not ported.

So today, reporting a canon quest persists completion but grants **nothing**.

## What "done" requires

1. Add a `rewards` schema to `data/quests/*.json` (per-difficulty `{item, quantity, meseta}`, or whatever shape the canon system wants — note canon difficulty keys are `normal`/`hard`/`super-hard`).
2. Populate rewards for the canon quests (the 16-quest spine + side quests).
3. Grant them on `report_quest` (SessionManager) / at the guild counter turn-in, into Inventory + meseta.
4. Characterization test: reporting a quest with defined rewards grants exactly those items/meseta.

## Why separate from #281

#281 removes the dead legacy mission apparatus (incl. its reward code). This is the *forward* feature — wiring rewards into the canon quest system — and is a **Beta save/reward deliverable**, not cleanup. Filing separately so #281 stays a pure deletion.

🤖 Filed by Claude Code (surfaced while scoping #281).

PR #353, opened 2026-06-13
feat(progression): difficulty unlock loop — story clear gates Hard/Super-Hard (#344)

### LINKED ISSUE #344: Difficulty unlock loop: story clear gates Hard, Hard clear gates Super-Hard
## Gap

Per-difficulty data exists everywhere — enemy scaling (enemy_spawner), technique levels, quest rewards (#331's normal/hard/super-hard tiers) — but **nothing gates or unlocks difficulties**. The guild counter's difficulty submenu presumably offers all three from the start. The clear→unlock loop is PSZ's core replay structure and the reason the per-difficulty reward tiers exist.

## Proposed scope (spec page first)
1. **Unlock state**: per-character `unlocked_difficulties` (CharacterManager persistence, like completed_missions); story clear (dark_castle reported) on Normal unlocks Hard; on Hard unlocks Super-Hard.
2. **Counter gating**: difficulty submenu greys locked tiers with the unlock condition (same locked-row pattern the quest list uses — "Complete the story on Normal").
3. **Audit**: confirm enemy stats/drops actually consume the difficulty everywhere (spawner does; check drop tables + boss HP).

## Tests
Unit: unlock state transitions on report of dark_castle per difficulty; counter entry availability. Post-build: matrix dc phase asserts hard unlocked in the post-dc save snapshot.

Small, self-contained, no dependencies — but it changes what the matrix's post-dc save looks like, so land before any Hard-mode content work.

🤖 Filed by Claude Code (beta coverage gap analysis).

kion-dgl:
## Why

Gate-lane Beta feature #3 — and the **0.35.0 "beta begins" bump**. PSZ's replay structure: every per-difficulty system already exists (enemy scaling, reward tiers from #318, technique levels) but nothing gated access to them. This is the loop.

## The rule (spec /states/difficulty-unlock)

- **Normal** always unlocked.
- Reporting the story finale (`dark_castle`) on **Normal** → unlocks **Hard**; on **Hard** → **Super-Hard**. Only the finale gates; side/variant quests don't.

## What

- **GameState**: `unlocked_difficulties` (mirrors `completed_missions` exactly), `is_difficulty_unlocked` / `unlock_difficulty` / `apply_quest_clear_unlock` + the progression constants.
- **CharacterManager**: per-slot swap (set_active_slot + sync) and a **v7 seed migration** — a character whose `completed_missions` already has the finale retroactively keeps Hard (old saves can't recover the cleared difficulty, so conservative: first tier up only).
- **SaveManager**: save v7 + migration hook.
- **guild_counter**: locked tiers render muted with `[LOCKED]` + an unlock-condition hint; accept is blocked at *both* the input gate and the async confirm path; turn-in shows "★ Hard difficulty unlocked!". Extracted `_render_difficulty_submenu` to keep `_refresh_display` under the complexity bound — **the ratchet caught the regression and I refactored rather than baselining it** (per /engineering).

## Two test layers (per /engineering)

- **Unit** (`test_difficulty_unlock`): the finale rule per tier, always-on Normal, non-finale no-op, per-character persistence through the slot swap, and the v7 migration.
- **Post-build**: the regression matrix's `dark_castle` phase now asserts Hard is unlocked in the post-dc `save_data.json` (jq), flipping the run to a loud failure if the wiring doesn't fire (the #335 integration lesson — verified the jq logic both directions).

## Verification

- Suite **1620 passed, 0 failed** | ratchet 0 new (complexity back to 69) | spec builds 96 pages | matrix script syntax + jq checked
- ⏳ **Full regression matrix running** (it changes the post-dc save + adds the assertion) — results before merge

The autopilot always picks difficulty index 0 (Normal, always unlocked), so the gating doesn't perturb the matrix.

Closes #344.


kion-dgl:
## ✅ Matrix re-run green — chokepoint fix verified

After moving the unlock into `SessionManager.report_quest` (the chokepoint both report UIs call), the post-build assertion passes:

- **16/16 phases pass** (boot → dark_castle → shops)
- **post-dc save: `unlocked_difficulties = ["normal", "hard"]`** — Hard unlocked by clearing the finale on Normal, persisted to disk (matrix exit 0, no MATRIX_FAIL)

This is the two-layer policy working exactly as intended: the first run's post-build assertion caught that hooking only the guild counter missed the autopilot's city-office report path (a #335-class integration gap the unit test couldn't see), and the fix is verified in the running game.

Proceeding to the sanity merge gate.

PR #355, opened 2026-06-13
feat(quests): report at the guild counter only — remove Principal report path (#354)

### LINKED ISSUE #354: Quest reporting belongs at the guild counter only — remove the Principal report path
## Design rule (from Kion)

> You should only be able to **accept** quests and **report** them to the guild counter. You SHOULD NOT be able to report quests back to the Principal to complete them.

So: the **guild counter** is the sole accept + report surface. The **Principal** (city office) is for story dialog only — intro and mission briefings — never quest turn-in.

## Current state (the bug)

There are **two** report paths:
- `scripts/2d/guild_counter.gd` — `has_completed_quest()` surfaces a "Report: <quest>" entry → `_open_report_modal` → confirm → `_report_quest`. **This is the canonical one and already works.**
- `scripts/3d/city/city_office_controller.gd` — `_on_principal_interact` calls `_report_quest` when `has_completed_quest()`. **This is the path to remove.**

The **autopilot** currently reports at the Principal (`_drive_office_report` → `PRINCIPAL_INTERACT_POS` → interact → `_on_principal_interact`), so removing the office path requires repointing it.

## Scope (one gate-lane PR)

1. **city_office_controller**: drop the `has_completed_quest()` → `_report_quest` branch from `_on_principal_interact`; remove the office `_report_quest` (and the #344 unlock dialog line that lives in it — the unlock stays in `SessionManager.report_quest`, which the guild counter calls, so it's unaffected). Intro/briefing dialog paths stay.
2. **autopilot — test BOTH surfaces (Kion's refinement)**: don't just repoint reporting to the counter. Instead, post-quest the autopilot:
   1. Goes to the office and **talks to the Principal** — the interaction must still *work* (dialog), so we keep exercising it.
   2. **Asserts the quest is NOT completed by talking to the Principal** — `has_completed_quest()` MUST still be true afterward. If talking to the Principal ends/reports the quest, that's a **regression → fail the phase**. This is the negative guard baked into the happy path.
   3. **Leaves the office**, goes to the guild counter, and drives the Report entry there (it's at index 0 when a quest is complete → ui_accept opens the report modal → ui_accept confirms — mirrors the existing `_drive_guild_counter` accept spam). `has_completed_quest()` flips false → `DONE ok`.

   Net effect: every quest phase now proves *both* that the Principal can be talked to AND that it cannot complete quests, plus that the guild counter does. A future regression that lets the Principal complete a quest fails the matrix automatically.
3. **spec**: make "reporting happens at the guild counter only" explicit on /states/story-progression (and the quest-counter shop page).

## Verification

- Full regression matrix (the report path is on the critical chain — every quest phase turns in). The #344 post-dc assertion is a bonus guard: it'll catch if the autopilot-report rewrite breaks the unlock, since the unlock rides `report_quest` regardless of which UI calls it.
- Unit: the office controller no longer reports (guard the `_on_principal_interact` branch is gone); guild counter report path already covered.

Lane: **gate** (flow + autopilot, no feel component). Best landed right after #344 (which the matrix is verifying now — its unlock architecture is correct under this design too).

🤖 Filed by Claude Code (Kion design directive).

kion-dgl:
## Why

Design rule (Kion): quests are **accepted AND reported at the guild counter only**. The Principal (city office) is story dialog — intro and mission briefing — never quest completion. There were two report paths; the Principal one is removed.

## What

- **city_office_controller**: dropped the `_report_quest` path entirely (and the #344 unlock dialog line that lived in it — the unlock stays in `SessionManager.report_quest`, which the guild counter calls, so it's unaffected). Talking to the Principal with a completed-but-unreported quest now shows a *"report at the guild counter"* dialog and does **not** complete the quest.
- **autopilot — both-surfaces test (Kion's design)**: post-quest the autopilot (1) talks to the Principal and **asserts the quest is still pending** — if the Principal ever completes a quest, `[sanity] FAIL` fires and the phase fails (negative regression guard, baked into the happy path); (2) leaves to the guild counter and reports there. `_drive_office_report` is now the Principal-guard; new `_drive_counter_report` + `_drive_guild_report` drive the counter turn-in.
- **spec /states/story-progression**: reporting is guild-counter-only; the Principal MUST NOT complete a quest.

## The clever part

Every quest phase now proves *both* that the Principal can be talked to AND that it cannot complete quests AND that the counter does — for free, on the existing critical chain. A future regression that lets the Principal complete a quest fails the matrix automatically. (And #344's post-dc unlock assertion now also guards that this report-path rewrite didn't break the unlock — the unlock rides `report_quest` regardless of which UI calls it.)

## Verification

- Suite **1622 passed, 0 failed** | ratchet 0 new | 168 scripts parse | spec builds 96 pages
- ⏳ **Full regression matrix running** — the report path is on every quest phase's critical chain, so this is the real test; results before merge.

Closes #354.


kion-dgl:
## ✅ Matrix re-run green (hang fix verified)

After the counter-report retry hardening + the corrected sanity oracle:
- **16/16 phases pass** (boot → dark_castle → shops)
- **#354 both-surfaces flow confirmed** across every quest phase: Principal-guard (does NOT complete) → guild-counter report → DONE
- **#344 post-dc assertion holds** — post-dc save `unlocked_difficulties = ["normal","hard"]` (verified directly; matrix stdout buffered to an empty log, but the report + save are ground truth)

The original hang was the no-retry counter-report interact + a concurrent zombie test_runner, both fixed. Proceeding to the sanity merge gate.

PR #365, opened 2026-06-13
feat(office): procedural library room for the Principal's office (#356)

kion-dgl:
## Why

Replaces the old in-house office GLB (a round chamber + long hallway) with a **procedural "library" room** built in GDScript — the look approved in the R3F screenshot loop (`web/src/office-editor/LibrarySet.tsx`, issue #356 reference images).

## The room (`scripts/3d/city/office_library_room.gd`)

A GDScript port of the approved `LibrarySet`:
- circular **tile1b** stone floor + dark border ring + ornate **floor05** medallion under the desk
- a **curved apse wall** (wpwall03) wrapping the back 270°, open at the +Z entrance (SurfaceTool arc)
- **fluted columns** (set02) flanking an **emissive sun stained-glass window** at the back-centre
- **bookshelves** along the arc (CC0 wood + colored procedural book spines)
- a **wood desk on a low stone dais** at the room centre
- warm scholarly key light + a cool window glow

Built via **composition** (`OfficeLibraryRoom.build(self)`), not scene inheritance, per the Android-export rule. Pack textures `load()` at runtime (the office was already pack-dependent via the principal GLB); the build falls back to flat stone if a texture is missing, so the script parses in repo-only CI.

## Layout + autopilot recalibration

Matched to the R3F `DEFAULT_LAYOUT` (R=9 hall, no hallway): spawn z=4.4 facing the window, door trigger z=6.5, Principal z=-5.6, desk/dais z=-4.5, quest-client slots pos_1/pos_2.

**Critically** (the #335/#354 lesson — the autopilot reads hardcoded office positions): `OFFICE_EXIT_POS` 27→6.5 and `PRINCIPAL_INTERACT_POS` -8.5→-3.9 (desk front, inside the 2.0 `INTERACTION_RADIUS`).

## CC0 vendoring (no upload / no pack republish)

ambientCG Wood062/Carpet013 (public domain) are vendored like the kenney packs: un-ignored in `.gitignore`, added to `check_asset_refs` in-repo prefixes, **excluded** from the R2 mirror (`check_r2_mirror` + `sync_tree` VENDORED_SKIPS), and added to the export `include_filter`. So `r2-mirror-check` won't probe them and no `sync-tree` is required.

## Verification
- Suite **1647 / 0** (170 scripts parse clean), `check_asset_refs` ok.
[...]

Issue #352, opened 2026-06-13
Charge cancel: expand drop options beyond dodge (N/H/S attack + menu open)


kion-dgl:
## Source

Playtest feedback from Rozalin on #273 (which only made **dodge** cancel a mid-charge technique). In PSZ you can drop a charge several ways depending on the situation, and forcing dodge-only breaks combat flow when you're already in melee range.

> In this clip I'm holding Heavy Attack to charge and canceling with a Normal Attack swing. With PSZ Godot, the equivalent would be holding the dedicated charge button and Normal/Heavy/Special would cancel it so you don't *only* rely on dodge (which forces a reposition). Since Start Menu and Quick Menu drop inputs, they too cancel it.

Rozalin's situational map:
- Charged PA → enemy attacking → **dodge** (already works, #273)
- Charged PA → enemy moved close → **Normal/Heavy/Special attack** (cancel + swing in place)
- Charged PA → enemies dead → **Start Menu** (drop without an action, keep moving)
- Charged PA → situation changed → **Quick Menu** (same)

## Current behavior

`_drop_charge` only fires inside `transition_to` for DODGING/DAMAGED/DOWN. Pressing another palette slot mid-charge does **not** release the active charge — if the new slot is a tech it overwrites `_charging_slot` (orphaning the first charge's visual); if it's an attack, `_execute_palette_action` runs but `_charging_slot` dangles.

## Proposed scope (one PR)

1. **Extract `_drop_charge()`** from the `transition_to` block (clear slot/tech/timer/ready, end visual, emit `tech_charge_released`) so every path shares it.
2. **N/H/S attack cancels**: in the palette-input path, when `_charging_slot >= 0` and the player triggers a *different* action that resolves to attack / strong_attack / special, call `_drop_charge()` first, then perform the attack. (A press on the *same* charging slot is still the normal release → cast.)
3. **Menu-open cancels**: opening the Start Menu or Quick Menu (#141) calls `_drop_charge()` — they drop inputs, so the charge dies silently with no action.

## Design note

PSZ has a *dedicated* charge button; ours charges by holding a palette tech slot. The cancel-options request applies either way, but if a dedicated charge bind lands later (alongside #141/#126), revisit so N/H/S vs. charge-slot disambiguation stays clean.

## Tests (two layers, per /engineering)

[...]

Issue #141, opened 2026-04-18
Quick menu behavior spec: pages, details panel, ordering, buffering


kion-dgl:
## Background

From design doc by Rozalin (2026-04-18). Quick menu is the lower-left-quadrant overlay for spontaneous weapon swap, item use, and tech casting. Minimal screen footprint, combat-friendly. Pairs with the start menu spec (sibling issue) as a core PSO "feel" element.

## Open + pages

- **Controller**: R2 + Triangle (back palette + interact/pickup combo)
- **Keyboard**: Ctrl + End
- **Default page**: weapon list
- **R2** flips page right: Weapon -> Item -> Technique

## Weapon list

- Single-press selects and closes the menu.
- **Circle** (PSO) / **Tab** (KB) / **Square** (PSZ) opens a details panel to the right of the list.
- Details panel has 3 sub-pages, flipped via L2/R2:
  1. Stat difference vs. currently equipped
  2. Weapon attribute
  3. Weapon description

## Item list

- Single-press selects and closes the menu.
- Circle does nothing.

## Technique list

- Single-press casts the tech and closes the menu.
- **Circle** starts a blinking highlight on the selected technique; selecting another slot swaps their positions (reorder).
- **Buffering**: recasting from quick menu mid-animation buffers the next cast -> frame-perfect chaining. Hotbar/palette casts do NOT buffer. Intentional - the behavioral variety is a skill-expression lever for experienced Force players.
- Reference: F3 technique menu also buffers and doesn't close on cast; experienced players switch between the two.

## List ordering

- Weapon and item lists match inventory order.
- **Clarification needed** before auto-sort lands: PSO default is unsorted with manual reorder + auto-sort toggle. Determine PSO's auto-sort rule (item ID? rarity?) before implementing.

## Last-page persistence

- Remembers last open page (weapon/item/tech) across field <-> Pioneer 2 transitions.
- On BB, persists even through room remakes; resets only on block change / logoff.
- For PSZ: interpret as "persists until return to title or close the game."
- Use case: switch to tech or item list in advance so it opens on the right tab when you need it mid-fight.

## Pioneer 2 / City behavior

- **Pioneer 2**: R2 + Triangle does nothing.
- **BB Lobby**: opens emote list; R2 toggles gender-variant emotes.

### Low-priority follow-up

- Allow emote list in Dairon City via quick menu (PSP2i permits this in Clad 6).
[...]

PR #366, opened 2026-06-13
feat(combat): expand charge-cancel to N/H/S attack + menu open (#352)

kion-dgl:
## Why

#273 made only a **dodge** cancel a mid-charge technique — which forces a reposition when you're already in melee range. Per Rozalin's playtest, PSZ lets you drop a charge several ways depending on the situation:

- Charged PA → enemy attacking → **dodge** (already works, #273)
- Charged PA → enemy close → **N/H/S attack** (cancel + swing in place)
- Charged PA → enemies dead → **Start Menu** (drop, keep moving)
- Charged PA → situation changed → **Quick Menu** (#141, same)

## What

- **Extract `_drop_charge()`** from the `transition_to` block — one shared cancel (clears slot/tech/timer/ready, ends the visual, emits `tech_charge_released`, no-ops when idle). The dodge/damage/down transitions (#273) now call it.
- **`_on_palette_pressed`**: a *different* palette input mid-charge drops the current charge first — an attack (N/H/S) cancels-and-swings in place; a different tech starts fresh with no orphaned visual. Re-pressing the **same** slot is untouched, so #273's hold→release cast still owns it.
- **Menu open**: `PsoStartMenu` gains an `opened` signal (there was no project-wide menu-open signal — `mobile_controls` polls for this reason); the player connects it to `_drop_charge`. Quick Menu (#141) should connect here too once it lands.

## Tests
- `test_charge_drop_paths`: `_drop_charge` clears + emits once + no-ops when idle; a different-slot press drops the old charge and starts the new one; the same-slot re-press does **not** drop (regression guard on #273's cast path).
- Suite **1655/0**; the existing #273 dodge-cancel test still passes (extraction preserved behaviour).
- Full regression matrix + sanity: running.

## Lane: **playtest** (combat feel)

The real acceptance is *feeling* the mid-combat cancel, so a bundle-signed sideload APK is attached for hands-on testing. Feel tuning (timing/animation-cancel windows) → follow-up issue if needed.

## ⚠️ Merge order

Version **0.36.1**, which assumes **#365 (0.36.0) merges first**. If this merges before #365, #365 will need a re-bump to 0.36.2 (version-check requires VERSION > main).


kion-dgl:
**Gates green — leaving open for your playtest.**

- Full regression matrix: **16/16 pass, 0 fails** (the charge/menu change touches `player.gd` + `pso_start_menu.gd`; no quest-chain regression).
- `sanity-check`: **28/28 pass**. Suite **1655/0** — incl. the new `test_charge_drop_paths` AND the existing #273 dodge-cancel test still passing (the `_drop_charge` extraction preserved behaviour).
- **APK**: `build/android/psz-charge-cancel.apk` — release-signed with the bundle keystore (build #111), so it installs in-place on the Pixel 8a / Retroid without an uninstall.

Playtest lane: this is a combat-*feel* change, so I'm not auto-merging — drop a charge with a Normal/Heavy/Special swing and with the Start Menu, see if the cancel timing feels right. Feel tuning (animation-cancel windows) → follow-up issue if needed.

⚠️ Version **0.36.1** — merge **#365 (0.36.0) first**, else this will need a re-bump to 0.36.2 (version-check requires VERSION > main).

PR #388, opened 2026-06-14
feat(shops): unified shop UI — Godot implementation of the spec mock (#368)

kion-dgl:
Per Rozalin's consistency feedback (#368) — the storage list diverged from the shop rows:

- Equipped gear marked with an **[E] prefix** instead of a right-aligned tag.
- **Rarity stars** move out of the row name into the **details panel** (a gold rarity line above the stats), matching how shops surface rarity.
- (Weapon-stats vs item-effect+count split already landed in #387.)

Mock only (Astro spec reference for the unified shop UI). Build green (97 pages). Version 0.36.7.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Issue #368, opened 2026-06-13
Unify the shop / counter interface into one common UI


kion-dgl:
From the onboarding playtest dump (#367). The item shop, weapon shop, Tekker (grind), photon collector, synthesis, and the storage/guild counters all reimplement the same list+detail layout inconsistently. They should share **one common shop/counter interface**.

Common contract the unified UI should enforce:
- **Detail panel (top-right) is the single source of item info.** It shows: stats, armor slot count, synthesis materials needed + how many you own, weapon stats, and *why* something is unavailable (can't afford / can't equip). None of this belongs inline on the list rows.
- **Disabled styling is consistent across every shop.** Today item-shop / weapon-shop / units use different "disabled" treatments. Pick one.
- **Class-can't-equip is NOT a buy blocker.** You should be able to buy gear your class can't equip; only **can't-afford** disables purchase (matches the #357 instance-equip philosophy).
- **Meseta readability.** Currently yellow on a light-blue panel — hard to read. Make it high-contrast and move it somewhere obvious (suggestion: bottom-left over the portrait).

Per-shop observations from the dump to fold in:
- Item shop: "can't afford" reason shown inline → move to detail panel; details panel is plain text, low visual appeal.
- Weapon shop: "can't use" (class) + "can't afford" inline on rows → detail panel only; armor slot count inline → detail panel.
- Units: disabled styling differs from item shop.
- Photon collector: shows red when you can't afford but still lets you attempt the buy (then errors) — should disable the buy, keep info hoverable.
- Synthesis: materials listed inside the item description → detail panel (materials, owned counts, resulting stats).

Likely needs design input on the exact panel layout. Source: #367.

kion-dgl:
The Astro spec mock (`/shop/weapon-shop` etc., `spec/src/components/screens/shops/` on the shared `ShopScreen`) is already close to the target unified interface — it's the working reference for this. First gap closed in #376: the item-details panel had an overflow scroll (hard-capped at 190px); it now flex-fills the column so details show in full. Remaining unification work (consistent disabled styling, class-can't-equip still buyable, meseta readability) continues against this spec.

kion-dgl:
[...]

PR #389, opened 2026-06-14
docs(claude): feature flow (spec-first) + sanity/matrix are Godot-only

kion-dgl:
Codifies two project rules (from Kion's workflow):

**Feature flow — spec-first, Godot-implemented, human-tested:** define the behavior in Astro first (a `/states` page or a shop/screen **mock** is the contract) → implement in Godot → add unit + autopilot tests → PR + matrix → Kion builds/tests on his Mac before it lands. A spec/mock-only PR (e.g. #388) is legitimate and lands as the *definition*; the Godot implementation that makes the runtime match is the next PR — not an incomplete one.

**Sanity/matrix are Godot-only:** run them only when the change touches `.gd` / scenes / quest data. Spec-only, docs, or `CLAUDE.md` PRs don't need the godot autopilot (CI `spec` build + `version-check` cover them) — running it just burns ~3 min and can flake on a loaded box.

Docs-only — per the second rule, **no godot sanity run** for this PR.

⚠️ Merge order: 0.36.8 assumes #388 (0.36.7) merges first.


kion-dgl:
Consolidated into #388 (its branch was cut from this one, so #388 now carries both the storage mock and the CLAUDE.md feature-flow rules). Closing to keep a single open PR; the Godot unified-shop implementation (#368) continues on #388's branch.

PR #434, opened 2026-06-28
fix(city): bound walkable floor with perimeter walls (#371)

### LINKED ISSUE #371: City areas: player walks off the floor into the void (no boundaries)
From the playtest dump (#367). In the counter area you can walk off the visible floor into empty space.

The floor *collider* exists (\`city_counter_controller.gd:67\`, a 30×50 box at y=0) but there are **no boundary walls/limits**, so the player keeps walking on the invisible collider past the visible geometry. Needs perimeter collision (walls or a walkable-area limit), and ideally a kill-z / respawn fallback.

⚠️ The new procedural Principal's office (#356/#365) uses the same flat-box-collider-no-walls pattern (24×24) — it likely shares this and should be fixed together.

---

## Direction update — 2026-06-29 (@kion-dgl): trace a real floor polygon, not a bounding box

**PR #434 (bounded-collider + perimeter walls) is closed — the approach doesn't hold.** The five city areas (market, sewer/underground, counter, teleport/warp, principal's office) don't expose usable floor triangles to derive a floor mesh from (unlike the field areas). #434 approximated each with a flat `y=0` slab + a hand-sized rectangle (cylinder for the office) of perimeter walls — but rectangular boundaries don't match the irregular floor shapes: you still walk past the painted floor in the corners and get stopped in the wrong places. **A bounding box pretending to be a floor.**

| Approach | Verdict |
|---|---|
| Flat `y=0` box collider, no walls (current `city_*_controller.gd`) | ❌ walk off into the void |
| #434: slab + rectangular/cylindrical perimeter walls per area | ❌ closed — boundary shape ≠ actual irregular floor |
| **Traced per-area floor polygon → walkable mesh/boundary** | ✅ chosen path |

### Chosen path: a floor-tracing tool
Build a small **three.js tool** that loads each city area's existing geometry as a visual reference, lets the dev **place points on the floor**, and emits a **polygon per area**. That traced outline becomes the actual walkable floor mesh / boundary in Godot (mirrors the original three.js "walk on `y=0` + walls", but with a real per-area outline instead of an approximation). **Next step is the tracing tool, not another collider-shape tweak.**

```mermaid
flowchart LR
    geo[city area geometry\n(visual reference)] --> tool[three.js floor-tracing tool\nplace points on floor]
    tool --> poly[per-area floor polygon]
    poly --> godot[walkable floor mesh / boundary in Godot]
    godot --> clamp[player confined to traced outline\nno void, no false stops]
```

### Expected behavior (RFC-2119 spec contract)
- Each of the five city areas (market, sewer/underground, counter, teleport/warp, principal's office) **MUST** have a walkable boundary derived from a **traced floor polygon** matching its visible floor — **MUST NOT** rely on a rectangular/cylindrical bounding approximation.
- The player **MUST NOT** be able to walk off the visible floor into empty space, and **MUST NOT** be stopped short inside the visible floor (no false walls in the corners).
- The tracing tool **MUST** emit a per-area polygon in a stable format consumable by the Godot floor/boundary builder (one outline per area; closed loop).
- A **kill-z / respawn fallback** **SHOULD** remain as a safety net if the player ever escapes the boundary.

### Likely code / artifacts
- Per-area controllers with the flat box collider: `city_counter_controller.gd:67` (30×50), the procedural office (#356/#365, 24×24), and the other three city controllers — all replace the box with the traced-polygon boundary.
- New: the three.js floor-tracing tool (web tooling) + a per-area polygon artifact checked into the repo; a Godot-side builder that turns the polygon into the walkable mesh/boundary.

### Tests (two-layer)
- **Unit:** with a traced polygon loaded for an area, assert a point just outside the outline is non-walkable and a point just inside is walkable (boundary matches the polygon, not a bbox).
- **Autopilot probe:** walk a city area toward each corner; assert the player is confined to the traced floor (no void fall, no premature stop).

kion-dgl:
## Root cause

City areas walked the avatar off the visible floor "into the void." The bug is an **honesty gap between the floor collider and the rendered floor mesh**:

`CityAreaBase._add_floor_collision` built a **single flat `BoxShape3D` slab** (top at `y=0`, extends downward, **no walls**) on environment collision layer 1. Callers passed a footprint *larger than / a different shape from* the visible floor:
- counter → `30×50` box (`city_counter_controller.gd:67`)
- office → `24×24` square (`city_office_controller.gd:119`) while `office_library_room.gd` only renders an **R=9 round disc**

The player's ledge-guard (`_can_move_to` / `_has_floor_at`, raycasting layer 1; documented in `spec/states/player-state` — "requires at least 2 of 3 to find floor … prevents walking off ledges") then **found this invisible overhang past the visible edge** and permitted movement onto it. Because the avatar never actually left the slab (it stayed at `y=0` on the oversized collider), the existing fall-respawn (`FALL_RESPAWN_Y=-10`, `player.gd`) **never fired** — the player just stood on invisible floor beyond the world geometry.

## Fix (why, not just what)

Make **"floor collider == walkable footprint == bounded by walls"** a single base-class invariant, so the fix can't drift per-area:

1. **`_add_floor_collision` erects an invisible perimeter by default** — a `FloorBoundary` `StaticBody3D` (layer 1 / mask 0) with four thin tall `BoxShape3D` walls hugging the floor edges. All five city areas (counter/market/warp/underground/office) gain containment with **zero call-site churn**; the player (layer 2, mask 1) is physically stopped at the edge.
2. **Round path for the office** — a new `radius>0` argument builds a `CylinderShape3D` floor + a ring of wall segments with a **gap at the +Z door**. The office switches to `R=9` so the avatar stays on the visible disc instead of the old `24×24` square.
3. **No doorway gets sealed** — exit triggers already sit *inside* the wall line (counter: market `z=+20`, warp `z=-22.3`, office `x=+11.5`, all within the `±15x/±25z` walls), so a transition fires *before* the avatar reaches a wall; the office door has an explicit +Z gap. `FALL_RESPAWN_Y` stays as the **secondary** safety net.

## Spec (spec-first, reconciled)

New normative page **`spec/hierarchies/city-area.astro`** — the `CityAreaBase` contract (none existed; engineering policy requires one page per behavior hierarchy and this base has 5 subclasses). RFC-2119: every area **MUST** enclose its walkable floor with perimeter collision on layer 1; footprint **SHOULD** match the rendered mesh (round rooms use a cylinder + ring with a door gap); exit volumes **MUST** sit inside the wall line; fall-respawn is the **secondary** net. Reconciled with `states/player-state` (does not contradict it — states that bounding walls make the ledge-guard *deterministic* by guaranteeing no invisible floor past the visible edge). Linked from `hierarchies/index`.

## Verification

**Unit test (added) — `test_city_area_floor_boundary` (`test_runner.gd`, registered in `_run_tests_systems`):** builds both paths, asserts the `FloorBoundary` body + layer 1 / mask 0, four rect walls at exactly `±(half-extent + thickness/2)`, the round `R=9` `CylinderShape3D` floor, and the `+Z` door gap (no wall across the entrance, back walled). Asset-free, exact-equality.
- Result: **all 15 assertions PASS.**

**Full suite (`godot --headless res://scripts/tools/test_runner.tscn`):**
- My branch: **1917 passed, 8 failed**
- Pristine `main` baseline (same command): **1907 passed, 9 failed**
- The remaining failures are the **pre-existing repo-only pack-preload `script_parse` failures** (`start_warp`, `companion_npc`, `waypoint`, `gate`, `valley_field_controller`, the minimaps, etc. — scripts that transitively `preload()` pack-only SEGA assets absent in a repo-only checkout; documented). They are byte-identical to `main` and **untouched by this PR** — my branch has *fewer* of them and 10 more passes. CI (with the pack/assets present) compiles these. **No regression introduced.**
- Note: the non-zero exit code is also present on pristine `main` for the same reason; it is not introduced here.

**Autopilot probe (`scripts/tools/autoplay/shop_smoke.sh`, local pack):** added a runtime `[sanity] checkpoint: office-boundary-built` in the office `_ready` (asserts the `FloorBoundary` node built at runtime) and an oracle line in `shop_smoke.sh`.
- Result: **25 passed, 0 failed, `[sanity] DONE ok`, rc=0.** The new checkpoint fired and the full office→shops→storage flow ran clean — confirming the round-floor swap did **not** break office spawn/navigation/NPC interaction/exits.

## NEEDS MAC VERIFICATION

- The **live containment-walk** for the **rectangular** areas (counter/market/warp/underground): drive the avatar into `+x`/`±z` past the visible edge and confirm `global_position` clamps at the wall. Headless movement physics is flaky under xvfb, and `shop_smoke` only loads the **office** (round path). The deterministic guarantee is covered by the unit test; this is runtime integration confidence on real hardware.
- Visual check that no perimeter wall is reachable/visible inside any area and that all four area exits still transition.

Honest caveat: rectangular areas' colliders may still slightly exceed their visible meshes; the walls cap walk-off at the collider edge, and precise per-area tightening to the visible edge is follow-up polish.

Closes #371


kion-dgl:
Closing — the bounded-collider approach here is broken in practice. A flat y=0 slab plus a hand-sized rectangle/cylinder of perimeter walls per area doesn't match the actual (irregular) floor shapes, so the player still walks past the visible floor and gets stopped in the wrong places. This needs a real traced floor polygon, not a bounding box. See #371 for the path forward.

PR #440, opened 2026-06-28
fix(elements): emulate MirroredRepeatWrapping via UV-fold; drop Godot-3 flags_mirrored_repeat (#381)

### LINKED ISSUE #381: Three.js→Godot port: emulate MirroredRepeatWrapping via UV-folding math
Reference / action item for porting Three.js shaders & materials to Godot (e.g. the office library port #356, and future stage/effect ports).

**Problem:** Godot has no native equivalent of Three.js `MirroredRepeatWrapping`. The coordinate folding must be done in shader math.

**Action:** when porting a Three.js shader to a `.gdshader`, replace standard UV sampling with the mirrored fold. Scale the UVs first, then fold:

```glsl
vec2 scaled_uv = uv * tiling;          // your repeat/scale factor
vec2 mirrored_uv = 1.0 - abs(mod(scaled_uv, 2.0) - 1.0);
// sample with mirrored_uv instead of scaled_uv
vec4 col = texture(albedo_texture, mirrored_uv);
```

This yields seamless mirror-tiling that matches the Three.js mock. Capture as the standard recipe so every shader port applies it. (Likely also belongs in the engineering spec / CLAUDE.md porting notes.)

kion-dgl:
## Why

#381 asks to capture the standard recipe for porting Three.js `MirroredRepeatWrapping` to Godot (no native equivalent) and to put it in the engineering spec. Grounding that in the real code surfaced that the issue's exact failure mode is **live in the trap elements today**.

`GameElement._setup_split_materials` (`scripts/3d/elements/game_element.gd`) wrote:

```gdscript
std_mat.albedo_texture.flags_mirrored_repeat = true
```

`flags_mirrored_repeat` is a **Godot 3** `Texture` flag — it **does not exist on Godot 4 `Texture2D`/`ImageTexture`**. In GDScript this assignment does not silently no-op; it raises at runtime:

```
Invalid assignment of property 'flags_mirrored_repeat' … on a base object of type 'ImageTexture'
```

A runtime error **aborts the enclosing lambda at that line**. So for every *textured* surface, the statements that follow inside the callback — `mesh.set_surface_override_material(surface, dup)` and the `result["feature"]/result["base"]` registration — never executed. `_setup_split_materials` returned `{feature: null, base: null}`.

Downstream, `bear_trap.gd` / `needle_trap.gd` store those as `_prong_material` / `_base_material` and guard every use with `if _prong_material:`. With both null:
- the prongs never hid on the `off` (triggered) state (`_apply_state`), and
- the held-laser texture never scrolled (`_update_animation`).

Reproduced in isolation: the throw aborts the callback and leaves the result dict untouched.

## The fix

Remove the bogus line. The duplicated `StandardMaterial3D` already carries the texture and nearest filter; the override now applies and feature/base resolve. These trap surfaces don't tile past `[0,1]`, so plain repeat is correct. Where mirror-tiling genuinely **is** needed, the right Godot-4 path is the existing UV-fold shader (`mirror_repeat.gdshader` via `_setup_mirror_textures`) — never a texture flag.

## Spec (the #381 deliverable)

New normative page **`/engineering/mirrored-repeat-wrapping`** (`spec/src/pages/engineering/mirrored-repeat-wrapping.astro`), linked from `/engineering`. It documents:
- the trap (Godot 4 has no native mirror flag; the Godot-3 flag throws and aborts setup),
- the UV-fold recipe `1 - abs(mod(uv*tiling, 2) - 1)` + the canonical `mirror_repeat.gdshader`, with a table of reference fold values matching Three.js,
- when **not** to use it: `RepeatWrapping` ports (e.g. `office_library_room.gd`, which mirrors `LibrarySet.tsx`'s `THREE.RepeatWrapping`) correctly stay `StandardMaterial3D` + `uv1_scale`,
- the animation-vs-shader tradeoff (surfaces needing `uv1_offset` scroll / alpha-scissor must stay `StandardMaterial3D`).

No existing spec contradicted; this is a new page.

## Tests

`test_runner.gd` → **`test_mirror_repeat_wrapping`** (one registration line) pins both layers of the recipe:
- (a) `_setup_split_materials` now resolves **non-null** feature/base for textured surfaces and applies the override to the mesh (the regression),
- (b) `_setup_mirror_textures` applies a `ShaderMaterial` bound to `mirror_repeat.gdshader` at the documented `uv_scale = (2,2)` with `mirror_x/mirror_y` enabled.

## Verification

- **New test logic: GREEN (8/8).** Verified against the real `GameElement` code with a standalone `SceneTree` harness (the same asserts the registered test makes): feature/base non-null, override applied, nearest filter, `ShaderMaterial` applied, shader loaded, `uv_scale (2,2)`, `mirror_x`. `rc=0`.
- **Root cause independently reproduced:** isolated scripts confirm `flags_mirrored_repeat` does not exist on Godot-4 `Texture2D`, that assigning it throws, and that the throw aborts the callback leaving the result dict untouched.
- **Baseline suite GREEN:** the clean `main` `test_runner.tscn` run is `1912 passed, 0 failed`, `157 scripts parsed cleanly`. This diff is additive (one test) and only *removes* a throwing line, so it cannot lower the baseline.
- **Could NOT run the full registered `test_runner.tscn` suite in this PR's worktree:** a fresh `git worktree` lacks pack-only assets that aren't in git (`assets/fonts/JetBrainsMono-*.ttf`, `assets/ui/stats.png`) which the full scene's theme pulls in, so the scene can't boot there (hangs before `_ready`). My diff touches none of those. Run the registered suite on the main checkout / Mac (where pack assets are present) to confirm `1921 passed, 0 failed`.
- **No autopilot probe run.** The fix is a runtime *visual* (trap prongs hiding + laser scroll); asserting it headless needs the valley quest + the trap GLB. The deterministic seeded unit test is the primary gate here; a visual confirmation belongs on a Mac run.

### NEEDS MAC VERIFICATION
Trigger a bear trap / needle trap in the valley quest and confirm: prongs **hide** when the trap flips to `off`, and the held-laser texture **scrolls** while a player is caught. Both were silently broken by the aborted material setup; this restores them. Also run the registered `test_runner.tscn` suite to confirm green with the new test.

Closes #381


kion-dgl:
Closing as a duplicate of #441 (same issue #381). The overnight workflow retried the implement agent and it opened two branches for this issue; #441 is the canonical one recorded by the run. Branch left intact if you want to compare the alternate UV-fold approach.

PR #436, opened 2026-06-28
fix(player): Start Menu captures the confirm press (modal > world > palette) — #426

kion-dgl:
## Why

The Start Menu could let a confirm press fall through to the world interaction underneath it: stand on the central **WarpTeleporter** pad, open the Start Menu, press confirm → the warp overlay opened *behind* the menu. The same structural gap also let one button bound to **both** a world interaction and a palette/free action fire **both** on a single press (the generalized #423 case).

### Root cause
`scripts/3d/player/player.gd` `_unhandled_input` ran the palette block and the interact block as **independent, non-consuming `if`s** — neither consumed the event nor returned after a world interaction fired, so there was no "world interaction wins and consumes" arbitration. The only thing blocking interact under an open menu was the coarse `is_gameplay_blocked()` early-return; `_try_interact()` itself fired unconditionally with no defense-in-depth.

Movement was already correct: it runs from the ungated `_physics_process`, so the player keeps walking with the menu open (confirmed by tester Sazonde99) — that behavior is preserved.

## The fix

Encode the precedence rule at the single point it is decided, behind a pure testable seam:

- **New pure static** `Player.arbitrate_field_actions(pressed, has_interactable, in_city) -> Array[String]` returns the ordered effects for one `InputEvent`: a world interaction (interact pressed **and** an interactable in range) returns `["interact"]` and **drops** any palette action on the same press; otherwise it returns the palette / dodge / `palette_swap` actions that matched (suppressed entirely in city).
- `_unhandled_input` builds the pressed-action set, calls the helper, and when the result is a world interaction fires `_try_interact()`, marks the event handled, and **returns before any palette dispatch**. Palette *releases* still fire (a release is not a competing press).
- **Defense-in-depth:** `_try_interact()` early-returns when `is_gameplay_blocked()`, so a world interaction can never fire under an open modal even via a future input path.

### Precedence (now spec'd, RFC-2119)
`spec/src/pages/states/start-menu.astro` gains a **Confirm / interact precedence** table:

| # | Condition | Outcome |
|---|---|---|
| 1 | modal / Start Menu open | modal consumes the press; world + palette MUST NOT fire; **movement stays enabled** |
| 2 | world-interaction prompt in range on `interact` | world interaction CONSUMES the press ahead of any palette action on the same button (#423) |
| 3 | neither | palette / free action fires (in-city combat still suppressed) |

[...]

Issue #426, opened 2026-06-28
Start Menu open should capture input — blocks simultaneous teleporter interaction


kion-dgl:
**Source:** [playtest report 2026-06-28T03-42-48](https://psz.onl/reports/2026-06-28T03-42-48-755Z-2ce90f9c) — Rozalin, v0.37.15

## Observed
> "With the Start Menu open, you shouldn't be able to simultaneously interact with the City Teleporter as the Start Menu should be taking your inputs instead."

Report log confirms `[Player] Interacted with: WarpTeleporter` fired while the Start Menu was open.

## Proposed spec
While the Start Menu is open, it MUST capture/consume **confirm/interaction** input:
- World-interaction prompts (teleporter, shop NPCs, pickups) MUST NOT trigger from the confirm action while the menu is open.
- **Player movement MUST remain enabled** — the Start Menu is a special menu that intentionally allows simultaneous **movement AND menu navigation** (tester Sazonde99). Do NOT freeze the player.
- Closing the menu MUST restore normal world interaction on the next frame, with no buffered confirm leaking through to a world interactable.

## Likely code
- `scripts/3d/field/pso_start_menu.gd` / `scripts/3d/field/start_menu_input.gd` — the menu uses `_unhandled_input`. The teleporter likely reads the confirm action directly (polling `Input.is_action_*`) rather than via `_unhandled_input`, so it never sees the menu consume the event. Gate world interactables on a "menu open" check, or have the teleporter/interaction read through the unhandled-input chain so an open menu can `set_input_as_handled()`. (Note: the menu must consume **confirm/interact**, not movement — movement stays live.)
- Interaction side: `scripts/2d/warp_teleporter.gd` and the shared interaction prompt handler.

## Tests (two-layer)
- **Unit:** open the start menu, dispatch a confirm press at a teleporter, assert no warp/interaction fired; dispatch a movement input, assert the player still moves.
- **Autopilot probe:** stand on a teleporter, open start menu, press confirm, assert no area change in a sanity run; assert movement still works with the menu open.

Relates to #141 (menu input/buffering behavior).

---

### Generalized precedence rule (added 2026-06-28, tester Sazonde99 via #423)
The same input-routing gap shows up even with **no Start Menu open**: when a single button is mapped to *both* a world interaction (teleporter/telepipe warp) *and* a palette/menu action, both fire on one press. Per tester feedback, the rule is:

[...]

PR #470, opened 2026-07-03
fix(#426): carried-open Start Menu loses its input block on area transition — fix + two-layer regression coverage

kion-dgl:
## Why

Rozalin's follow-up on #426: open the Start Menu in the market, leave it open, transition to the guild counter area, walk up to the counter — the interact fires **through** the open menu. The #436 close-out assumed the carried-over case was covered as a side effect of the player-side arbitration; it wasn't.

## Root cause

`SceneManager.goto_scene()` hard-reset `GameState.modal_stack = 0` on every area transition. That reset exists for a good reason (per-scene dialog boxes freed mid-dialog never pop their modal, wedging input in the next scene) — but `PsoStartMenu` is an **autoload** that survives the transition still open, so the reset silently dropped *its* live block too: menu visibly open, `is_gameplay_blocked()` false, next confirm/interact double-fires. The player-side gates (#436 `arbitrate_field_actions` + `_try_interact` early-return) were never the hole — they both check `is_gameplay_blocked()`, which had been zeroed out from under them.

## Fix

`goto_scene`'s reset is extracted into `SceneManager._reset_modals_for_scene_change()`: still zeroes leaked per-scene dialog modals, then **re-asserts the surviving menu's modal** when `PsoStartMenu.is_open()`.

## Coverage (two layers, red→green proven)

- **Unit** — `test_menu_carry_survives_scene_signal()`: menu open → `scene_changed` signal → still open + still blocked; then the extracted reset is called directly with a simulated leaked dialog modal on the stack — asserts the menu's block survives (this exact assertion was red before the fix) and the leak is still cleared (`modal_stack == 1`). Suite: 2058/0.
- **Autopilot probe** — `PSZ_AUTOPILOT_MENU_CARRY=1` (opt-in via `SANITY_MENU_CARRY=1` in `sanity_check.sh`; the default sanity run is byte-identical): opens the menu in the office, carries it across the office→counter transition, teleports to the guild NPC and presses interact — fails hard if the guild counter overlay opens, then closes the menu and resumes the normal flow so the rest of the run still completes. **Before the fix: reproduced Rozalin's double-fire exactly** (guild counter opened through the carried menu). After: `start-menu-carry-blocks-interact` checkpoint + full run green, 29/29 checks.
- **Spec** — `/states/start-menu` invariants gain the carried-over clause from the #426 edge case + acceptance criterion (precedence MUST be invariant to whether the menu was opened here or carried over).

Complements the fresh-open probe (`PSZ_AUTOPILOT_MENU_GATE=1`) which covers the same-area case at the warp pad.


kion-dgl:
Windows build for testing — built 2026-07-02 from `6fc4a732` (v0.39.1):

**Download:** [psz-godot-pr470.zip](https://psz.onl/pr-builds/psz-godot-pr470.zip) (121 MB)

**What to check (Rozalin's #426 repro):** open the Start Menu in the market (or any area), leave it open, transition to the guild counter area, walk up to the counter and press accept/interact — the press must land in the menu only; the guild counter must NOT open underneath it. Same for the warp teleporter pad. Closing the menu and pressing again should interact normally.

PR #488, opened 2026-07-04
Fix #477: big-rig enemies wedge in ATTACKING (freeze until hit)

### LINKED ISSUE #477: Large enemies freeze (AI + animation) on player collision until hit; small enemies unaffected
**Source:** [playtest report 2026-07-03T17-02-46](https://psz.onl/reports/2026-07-03T17-02-46-045Z-72afc3d4) — Rozalin, build 405.

| Enemy | Player runs into it | On hit |
|---|---|---|
| **Large** | AI + animation **freeze** ❌ | unfreezes (resumes) |
| Small | keeps moving/attacking ✅ | — |

The player physically colliding with a large enemy halts its `_physics_process` AI/animation until a damage event kicks it back into a live state.

## Likely cause (grounded)
`scripts/3d/enemies/enemy_base.gd` — the state-machine `_physics_process` (L314) drives AI + `move_and_slide()`, gated in part by `_is_immobilized` (L55, "frozen/stunned/sleeping"). A player↔large-enemy collision appears to wedge the enemy: either
- `move_and_slide()` on the large body stalls against the player's collider (bigger shape / higher mass) and the AI tick short-circuits, or
- contact flips the enemy into an immobilized/idle-ish state that only clears on the HURT transition (L359) — which is why an **attack unfreezes it**.
Smaller enemies don't hit the threshold, so they keep ticking.

- An enemy's AI and animation **MUST** continue running while the player is in physical contact with it, regardless of enemy size — collision **MUST NOT** freeze its state machine or `_physics_process` AI.
- Player↔enemy contact **MUST** resolve as pushback/overlap only; it **MUST NOT** set `_is_immobilized` or otherwise pause the enemy. An enemy **MUST NOT** require a damage event to resume behaving.
- Large and small enemies **MUST** behave identically under player contact (differ only by their authored stats/speed).

```mermaid
stateDiagram-v2
    Chasing --> Chasing: player runs into it (large) — MUST keep ticking
    note right of Chasing
        BUG: large enemy freezes on contact,
        only HURT (on hit) resumes it.
        Contact MUST NOT immobilize.
    end note
```

## Likely code / next steps
- `enemy_base.gd` `_physics_process` (L314–368): confirm whether a large-enemy collision zeroes velocity/short-circuits the AI, or sets `_is_immobilized`. Ensure the AI/animation tick runs independent of contact.
- Check collision layers/masks + body size between large and small enemy scenes for the asymmetry.

## Tests (two-layer)
- **Unit:** place the player overlapping a large enemy for N frames, assert `current_state` keeps advancing and `_is_immobilized` stays false (no attack applied).
- **Autopilot:** walk the player into a large enemy in a combat probe, assert its animation/AI keeps updating (no freeze) without hitting it.

kion-dgl:
Fixes #477.

## Root cause — not collision, and not `_is_immobilized`

The issue's original hypothesis (player collision stalling `move_and_slide` / setting `_is_immobilized`) turned out to be wrong. `_is_immobilized` is only ever set by status effects. The real mechanism:

- Walking into (or near) an enemy puts the player inside `attack_range`, so the enemy enters **ATTACKING** — that's the "on contact" trigger.
- `_process_attacking` spins until `is_attacking` is cleared, and the only clear path was `_on_animation_finished` parsing the *last underscore-suffix* of the finished clip and matching it against `"atk"`.
- The GLB rigs are prefixed by size (`s_` small / `m_` medium / `b_` big), and the **big rigs don't have a plain `atk` clip**: gorilla is `b_014_atk1`, roc `b_022_atk1` (both play via the `atk→atk1` alias, but finish as `"atk1"` — never matches), orangutan only has a misspelled `b_044_atckwat`, armadillo has no attack clip at all (those two play *nothing*, so the signal never fires). Also affected: seal `m_021_atk1`, snake `m_003_atk1`.
- Result: the enemy freezes in ATTACKING — zero velocity, one-shot clip stopped on its last frame — until `_on_hit_received` forcibly clears `is_attacking`. Exactly the playtest symptom: **large enemies freeze on contact and only unfreeze when hit; small enemies (all have `_atk` clips) are fine.**

## Fix (spec-first, per the working agreement)

1. **Spec** — `/states/enemies` gains an RFC-2119 *attack recovery* contract: ATTACKING MUST terminate for every rig regardless of clip naming/absence; attack end MUST be detected against the resolved clip actually played, never by suffix parsing; a damage event MUST NOT be the only recovery path; player contact MUST NOT freeze AI/animation.
2. **Godot** — `_play_animation` now returns the resolved full clip name; `_start_attack` records it (`_attack_anim`) or arms a fixed fallback duration when nothing resolves; `_on_animation_finished` clears by full-name match; the ATTACKING tick has an is-no-longer-playing watchdog. Bonus: the `atk` alias list gains orangutan's `atckwat` so that rig actually swings instead of timing out invisibly.
3. **Tests, two layers**:
   - `test_enemy_attack_recovery` (test_runner, 14 asserts): alias resolution for `atk1`/`atckwat` rigs, finished-signal clear by resolved name, watchdog recovery on a stopped clip, fallback-duration path for no-clip rigs, ATTACKING→LOAFING exits.
   - `PSZ_AUTOPILOT_ENEMY_FREEZE=1` autopilot probe: in the first field cell it spawns **Hildegigas** (the real gorilla rig) in contact with the player and requires the attack cycle to end with the player never swinging back.

## Verification (local, macOS headless)

| Check | Result |
|---|---|
| Full test_runner suite | **2179 passed, 0 failed** |
| Freeze probe on this branch | `DONE ok` — attack cycle ended (LOAFING) unhit |
| Freeze probe on pre-fix `enemy_base.gd` (red check) | `FAIL: wedged in ATTACKING (is_attacking=true anim_playing=false)` |
| `code_graph.py --check` / `orphan_files.py --check` | clean — dup and complexity baselines each net **−1** (probe-flag parsing extracted from `_ready`, fail helper reuses `_fail_and_quit`) |
| Spec Astro build | OK |

Note for the hardware pass: gorilla-family (Hildegigas/Hildeghana/Hildegao), roc-family (Pelcatraz/Pelcatobur), orangutan-family (Froutang/Frunaked), seal and snake families are the affected rigs — running into any of them should now show them keep attacking/loafing instead of freezing.

### DIFF
--- a/scripts/3d/enemies/enemy_base.gd
+++ b/scripts/3d/enemies/enemy_base.gd
@@ -44,6 +44,14 @@ const HURT_DURATION: float = 0.3
 ## Animation state tracking
 var is_attacking: bool = false
 var current_anim: String = ""
+## Resolved (full) name of the attack animation in flight — attack end is
+## detected against this, never by parsing a suffix out of the finished
+## signal (big rigs name the clip atk1/atckwat and used to wedge ATTACKING
+## until a hit forced HURT — #477, spec /states/enemies).
+var _attack_anim: String = ""
+## Ticks down when no attack animation resolved at all (armadillo has none).
+var _attack_fallback_timer: float = 0.0
+const ATTACK_FALLBACK_DURATION: float = 0.8
 
 ## Stuck detection — try perpendicular direction when blocked
 var _stuck_time: float = 0.0
@@ -508,11 +516,22 @@ func _process_chasing(_delta: float) -> void:
 		_face_direction(direction)
 
 
-func _process_attacking(_delta: float) -> void:
+func _process_attacking(delta: float) -> void:
 	velocity.x = 0
 	velocity.z = 0
 
-	# Wait for attack animation to finish, then loaf off
+	# Attack end: the resolved animation finished (signal path or the
+	# watchdog below), or the fallback duration elapsed when no attack
+	# animation resolved. Never wait on a damage event to recover (#477).
+	if is_attacking:
+		if _attack_anim.is_empty():
+			_attack_fallback_timer -= delta
+			if _attack_fallback_timer <= 0.0:
+				is_attacking = false
+		elif animation_player and not animation_player.is_playing():
+			is_attacking = false
+			_attack_anim = ""
+
 	if not is_attacking:
 		_start_loafing()
 
@@ -584,7 +603,10 @@ func _start_attack() -> void:
 		return
 
 	is_attacking = true
-	_play_animation("atk", true)  # Force play attack animation
+	_attack_anim = _play_animation("atk", true)  # Force play attack animation
+	if _attack_anim.is_empty():
+		# Rig has no resolvable attack clip — end the attack on a timer.
+		_attack_fallback_timer = ATTACK_FALLBACK_DURATION
 	_play_sfx("attack")
 
 	# Face the target
@@ -644,6 +666,7 @@ func _on_hit_received(raw_damage: int, _knockback: Vector3, accuracy: int = 100,
 
 		# Enter hurt state — play stagger animation, no physics knockback
 		is_attacking = false  # Cancel any attack
+		_attack_anim = ""
 		current_state = EnemyState.HURT
 		hurt_timer = HURT_DURATION
 		velocity = Vector3.ZERO  # Stop movement during stagger
@@ -711,14 +734,15 @@ func _has_floor_at(check_pos: Vector3) -> bool:
 	return not result.is_empty()
 
 
-## Play an animation by name (short name like "atk" will match "s_001_atk")
-func _play_animation(anim_name: String, force: bool = false) -> void:
+## Play an animation by name (short name like "atk" will match "s_001_atk").
+## Returns the resolved full animation name, or "" if nothing played.
+func _play_animation(anim_name: String, force: bool = false) -> String:
 	if not animation_player:
-		return
+		return ""
 
 	# Don't interrupt same animation unless forced
 	if not force and current_anim == anim_name:
-		return
+		return ""
 
 	# Try to find the animation - GLB animations are named like "s_001_atk"
 	var full_name := _find_animation(anim_name)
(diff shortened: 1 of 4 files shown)


kion-dgl:
🤖 **Autopilot note (triage loop):** attempted the `PSZ_AUTOPILOT_ENEMY_FREEZE=1` probe headless on this branch (isolated userdir, first-mission phase, local pack) — **inconclusive here, environment-limited, not a fault in this PR.**

The autopilot booted → char-select → city → guild counter fine, but then hit repeated `SCRIPT ERROR: Nonexistent function 'get_nearest_interactable' in base 'CharacterBody3D'` plus `player_rotation`/`fall_respawn_y` assignment errors on the player node. That's the signature of a **stale local pack** (my `dist/assets.pck` is from Jul 2, before the recent script changes): the packed `player` scene's script didn't attach, so the node fell back to bare `CharacterBody3D` and the guild interaction couldn't complete — the run never reached the first field cell where the freeze probe fires (`_on_field_cell_loaded`). So I couldn't exercise the Hildegigas-contact assertion locally.

Nothing here touches the player scene, so this is purely my sanity-pack drift. The evidence for the fix already stands on its own:
- unit layer `test_enemy_attack_recovery` (14 asserts) runs under CI's `test_runner` — no pack needed;
- your logged red→green on the probe itself (`FAIL: wedged in ATTACKING` on pre-fix vs `DONE ok` after);
- the diagnosis (big rigs finish as `atk1`/`atckwat`/no-clip so suffix-`atk` match never clears `is_attacking`) is consistent with the report symptom (large enemies freeze on contact, unfreeze only when hit).

Recommend the hardware pass run into the gorilla/roc/orangutan/seal/snake families per your note. If you want a clean local probe run, I'd need to rebuild/republish `dist/assets.pck` against current `main` first (didn't do that unprompted — no uploads without signoff). CI test-runner + spec build are green regardless.

Issue #239, opened 2026-05-27
Telepipe bugs: old pipe not deleted + quest counter locks to cancel after free-field return


kion-dgl:
### Bug 1: Duplicate telepipes
> new telepipe should delete the old one

Dropping a second telepipe doesn't remove the first. Expected: placing a new telepipe cancels/removes the previous one.

### Bug 2: Quest counter stuck on cancel after free-field telepipe return
> going to free field → telepipe → city, and interacting with quest counter locks you to cancel quest, making you stuck to whatever free field you're on

Steps to reproduce:
1. Enter a free field (no quest active)
2. Use a telepipe to return to the city
3. Talk to the quest counter (guild counter)
4. The only option is "Cancel Quest" — can't accept a new quest, effectively softlocked to the free field

kion-dgl:
## Scoping pass for Beta

**First Beta PR candidate** — it's a live playtest bug, no dependencies, no PSOBB reference needed. Two separate fixes:

1. **Duplicate telepipes**: TelepipeManager should cancel/free the previous pipe on placement. Characterization test: place → place → assert exactly one pipe node + the manager's state points at the new one.
2. **Quest counter stuck on cancel after free-field telepipe return**: likely a session-state flag (suspended session vs. accepted quest) not cleared on the free-field return path — reproduce via the sanity-style harness: free field → telepipe → return → open counter → assert the quest list renders instead of cancel-only. The guild counter's `_load_entries` branches on `has_accepted_quest() or has_suspended_session()` — a stale suspended session would lock it exactly like this.

Both fit one PR with the shop-smoke-style coverage; the autopilot already has telepipe steps in some phases.

🤖 Scoped by Claude Code.

PR #550, opened 2026-07-22
Quick Menu: capture input so palette/attack can't fire while open (#467)

### LINKED ISSUE #467: Quick Menu open should block action-palette / attack firing (input capture — #426 analog)
**Source:** [playtest report 2026-07-03T02-52-43](https://psz.onl/reports/2026-07-03T02-52-43-776Z-e812d2a0) — Rozalin, v0.39.0

While the **Quick (weapon) Menu** is open, action-palette actions bound to other buttons still fire. This is the Quick-Menu analog of the Start-Menu capture fixed in **#426** — the input-precedence arbitration doesn't treat an open Quick Menu as a modal that suppresses palette/gameplay actions.

- While the **Quick Menu is open**, it **MUST** consume/suppress action-palette and attack inputs — no palette action or attack **MUST** fire, regardless of which button it's bound to.
- Quick Menu navigation/confirm inputs **MUST** still work; closing the menu **MUST** restore palette firing on the next frame with no buffered press leaking through.
- This extends the #426 precedence: **active modal (Start Menu OR Quick Menu) > world interaction > palette/free action.**

```mermaid
flowchart TD
    press([action/attack button]) --> qm{Quick Menu open?}
    qm -- yes --> consume[Quick Menu consumes it\nno palette/attack fires] --> done([stop])
    qm -- no --> normal[normal palette / attack path] --> done
```

## Likely code
- `scripts/3d/player/player.gd` — `arbitrate_field_actions` / `_unhandled_input` (the #426/#458 precedence seam). It gates on `is_gameplay_blocked()` / the Start-Menu modal flag; add the Quick Menu open-state to that gate so palette/attack are suppressed while it's up.
- `scripts/3d/field/field_hud.gd` `_QuickWeaponMenu` exposes its open state (`opened` signal already wired at L100) — surface an `is_open`/GameState modal flag the arbitration can read.

## Tests (two-layer)
- **Unit:** open Quick Menu, dispatch a palette-bound attack press → assert no attack/palette action fired; close → assert it fires.
- **Autopilot:** open Quick Menu, press a mapped attack, assert no attack in a sanity run.
Relates: #141 (quick-menu spec), #426 (start-menu capture — same precedence family).

kion-dgl:
Fixes #467.

## Why
While the Quick (weapon) Menu is open, an attack bound to an action-palette slot on a button *other than* accept / cancel / `quick_weapon` still fired — the menu only consumed its own navigation actions and let every other press fall through to `Player._unhandled_input`. This is the Quick-Menu analog of the Start-Menu capture fixed in #426.

## What
Mirror the #426 pattern exactly: the Quick Menu now pushes a `GameState` modal on open and pops it on close, so `GameState.is_gameplay_blocked()` is true for the whole open window. The existing early-return on that gate in `Player._unhandled_input` (the same seam #426 relies on) is what suppresses palette/attack dispatch — no new arbitration logic.

Precedence is now the general rule the spec already anticipated: **active modal (Start Menu OR Quick Menu) > world interaction > palette / free action.**

**Movement stays live.** The modal gate only blocks the action path in `_unhandled_input`; movement runs from the ungated `_physics_process` (see the note at `player.gd` `_try_interact`), so the Quick Menu — like the Start Menu — still allows moving and navigating the list at the same time. The menu does **not** freeze the player.

Balance safety:
- `_open()` pushes the modal only *after* the non-empty-list early return, so an empty menu never opens and never blocks gameplay.
- `_close()` is idempotent (pops only if actually open), so a stray double-close can't underflow the modal stack and wrongly unblock gameplay under an unrelated modal.

## Spec
`spec/states/quick-weapon-menu` gains a normative **"Input capture (#467)"** section + invariant, and lists `game_state.gd`'s modal gate under "Implemented by".

## Tests (two-layer)
- **Unit** — `test_quick_weapon_menu_captures_input()` in `test_runner.gd`: open flips `is_gameplay_blocked()` true, close flips it back, a double-close does not underflow the stack under an unrelated modal, and an empty-list open pushes no modal. Full suite: **2310 passed, 0 failed**, warnings-as-errors gate clean.
- **Autopilot** — regression matrix run on this branch to confirm no gameplay regression (results posted below). A dedicated "open menu + press mapped attack → assert no attack" probe is a reasonable follow-up on the autopilot harness.

## Verification
Unit + autopilot on the minipc; final hardware pass on the Mac before merge.


kion-dgl:
## Autopilot: first-mission sanity ✅

Ran the Search & Rescue sanity playthrough on this branch (headless, `PSZ_AUTOPILOT=1`):

```
status: pass   godot_exit: 0   checkpoints: 99   duration: 397s
```

Boots → title → field → navigation → combat → quest clear (`[sanity] DONE ok`), with the new Quick-Menu modal gate in place. No regression to the normal input/field flow.

**Scope note:** this is a *regression* signal — it confirms the modal push/pop doesn't break normal gameplay. It does not specifically drive the #467 repro (open Quick Menu + press a palette-bound attack → assert no attack); that's covered by the unit test `test_quick_weapon_menu_captures_input()`, and a dedicated autopilot probe for it is a reasonable follow-up.

**Caveat:** run used the local Jul-6 asset pack (raw assets are on R2; can't rebuild a fresh pack on the minipc), so the definitive autopilot matrix is CI's PR playtest build + the hardware pass on the Mac.

Unit suite: **2310 passed, 0 failed**, warnings-as-errors gate clean.

kion-dgl:
## Full regression matrix ✅ — clean pass (rc=0)

Ran the complete autopilot regression chain on this branch (headless, 3× speed, local pack). **All 16 chained quests passed on the first attempt:**

| Quest | | Quest | |
|---|---|---|---|
| boot | ✅ | finding_ogi | ✅ |
| Search & Rescue | ✅ | static_in_the_snow | ✅ |
| the_paru_pact (canon) | ✅ | investigate_tower | ✅ |
| the_paru_pact (backtrack) | ✅ | heretic | ✅ |
| PP dialogue-bug probe (#239) | ✅ | control_system | ✅ |
| apothecary_supply (canon/moon/sol) | ✅✅✅ | the_broken_seal | ✅ |
| deep_ore_extraction | ✅ | dark_castle | ✅ |
| | | shops | ✅ |

Plus the **PP toast-persistence probe** (start menu during pickup): `toast-persistence OK — closed 'Picked up X' box stays hidden across start-menu toggles`.

> **Note on the `pp_menu_probe` sidecar:** its godot process hit the 480s budget (rc=124) at the slow end-of-quest warp-out, so the raw sidecar reads `fail`. But the matrix scores that probe by its toast-persistence assertion (which passed), so `MATRIX_FAIL` was never set and the run exited `rc=0`. It's also provably unrelated to this change — the Quick Weapon Menu is never opened in that probe (it exercises the **Start** menu); my modal push/pop only runs when the quick menu is actually opened.

**Verdict:** the #467 input-capture change has zero regression impact across the entire quest spine + menu/toast probes. Combined with the unit suite (2310 passed, 0 failed), this branch is ready for the hardware pass on the Mac.

_(Matrix ran against the Jul-6 local pack since raw assets are on R2; CI's PR playtest build re-runs autopilot against a freshly-built pack.)_

kion-dgl:
## Baseline confirmation — `pp_menu_probe` timeout is pre-existing

Ran the identical `pp_menu_probe` invocation (`the_paru_pact`, `PSZ_AUTOPILOT_MENU_DURING_PICKUP=1`, 3× speed, 480s budget) on a clean **origin/main** worktree (verified: 0 `push_modal` in `field_hud.gd` — no #467 change):

- **origin/main baseline → `rc=124` timeout** (stuck at title, 0 checkpoints).
- My branch (in the matrix) → also `rc=124`, but got further (past `quest_completed`) and its **toast-persistence assertion passed**.

Both time out on the same tight 480s budget, so the timeout **reproduces without this change** — it's a budget-tight/flaky probe, not a regression. My change also can't be the cause: the Quick Weapon Menu is never opened in this probe, so the `push_modal`/`pop_modal` code never runs (the baseline never even entered the quest).

**Net:** full matrix `rc=0` (16/16 chained quests pass, toast probe OK) + unit suite 2310/0 + baseline-confirmed pre-existing probe timeout. No regression attributable to #467.

PR #413, opened 2026-06-28
feat: player death/defeat flow + shop sell-tab ✕ markers

kion-dgl:
Two independent features bundled into one PR because they verify together in a single Mac build/test pass (one autopilot run + one shop walkthrough). Version bumped 0.37.16 → **0.38.0**.

---

### Why
Reaching 0 HP was a **soft-lock**: the avatar dropped into the `DOWN` lying-down loop and stayed there forever. A field session had no failure exit — only the success path (clear objectives → ride the end telepipe home). This adds the missing failure exit (classic PSO "you were defeated → return to the city").

### Behavior (spec-first)
New contract [`/states/player-death`](spec/src/pages/states/player-death.astro) (RFC-2119); [`/states/field-lifecycle`](spec/src/pages/states/field-lifecycle.astro) reconciled to document death as the session's failure exit.

- HP hits 0 in a field → knockdown + a translucent **red** overlay **below the HUD** (HP/PP stay visible on top); gameplay input blocked.
- PSZ-style box: **"You were defeated"** / **"Would you like to return to the city?"** with **Yes/No**. **No** re-shows the prompt; **Yes** is the only exit.
- **Yes** → city at the **telepipe-arrival spot** (no telepipe used/spawned), **session ends** (`return_to_city` — not resumable; cancels any telepipe), **revive to full HP**, **−50% carried meseta** (floored). **Banked meseta untouched.**

A setback, not a game-over: no permadeath, no item loss beyond the meseta penalty.

### Implementation
`GameState.lose_meseta()` · `SessionManager.defeat_return_to_city()` (the testable transaction) · `player.gd` emits `died` once on the HP-zero branch · new `defeat_screen.gd` (CanvasLayer @190, below FieldHud's 200) · `valley_field_controller` connects `player.died` → raises the screen.

### Tests (two-layer)
- **Unit** (`test_player_defeat_return`): meseta 100→50 / 101→51 / 0→0, bank untouched, revive, session-end, telepipe cleared.
- **Autopilot** (`PSZ_AUTOPILOT_DEFEAT` probe + `record_defeat.sh`): kills the player in the first field cell, confirms **Yes**, asserts city arrival with 50 meseta + full HP. Verified locally end-to-end → `DONE ok`.

---

### Why
[...]

PR #551, opened 2026-07-23
Fix #469: player damage-immune across the defeat→return-to-city window

### LINKED ISSUE #469: Player can still take damage during the defeat→respawn transition (should be invulnerable)
**Source:** [playtest report 2026-07-03T04-06-31](https://psz.onl/reports/2026-07-03T04-06-31-684Z-ee41cccb) — Rozalin, v0.39.0

Between pressing **Yes** on the defeat modal and the city load completing, the player's hurtbox is still live, so enemies can keep hitting the (already-defeated) player.

| Phase | Damageable today | Correct |
|---|---|---|
| HP hits 0 → defeat modal up | yes ❌ | no |
| Press **Yes** → transition to city | yes ❌ | no |
| In Dairon City | n/a (no combat) | no |

- From the moment the player is defeated (HP 0 / `died` emitted) through the return-to-city transition completing, the player **MUST** be damage-immune — its hurtbox **MUST NOT** register hits and no HP/PP change **MUST** occur.
- Immunity **MUST** cover the whole defeat window (modal shown, Yes pressed, scene transition), not just the final frame.

```mermaid
stateDiagram-v2
    Alive --> Defeated: HP 0 (died)
    Defeated --> Transitioning: Yes
    Transitioning --> City: load complete
    note right of Defeated
        MUST be damage-immune
        from here through City load
        (disable hurtbox / set invuln)
    end note
```

## Likely code
- `scripts/3d/player/player.gd` — on the HP-zero/`died` branch, set the player invulnerable / disable the attack-target hurtbox (the same hurtbox enemies read) for the defeat window; clear it after the city load.
- `scripts/3d/ui/defeat_screen.gd` / `SessionManager.defeat_return_to_city()` (#413) — ensure immunity persists across the transition, not just while the modal is up.

## Tests (two-layer)
- **Unit:** put player in defeated state, apply an enemy hit → assert HP unchanged and no further `died`/damage side effects.
- **Autopilot:** kill the player near a live enemy, confirm Yes, assert HP doesn't drop further during the transition (extend the `PSZ_AUTOPILOT_DEFEAT` probe).

Relates: #413 (defeat flow), #341 (death/recovery), #468 (death state persistence).

kion-dgl:
Closes #469.

## Why
On defeat, the player could still take damage between pressing **Yes** and the city load completing — enemies kept landing real hits on an already-defeated player during the transition. From the [playtest report](https://psz.onl/reports/2026-07-03T04-06-31-684Z-ee41cccb): *"When you are already dead and press Yes to respawn back to Dairon City, you can still take damage as you are respawning."*

Root cause is an **ordering bug**. The only "already dead" guard in `Player.take_damage()` was HP-coupled:

```gdscript
if current_state == PlayerState.DOWN and GameState.hp <= 0:
    return
```

But `SessionManager.defeat_return_to_city()` revives the player to **full HP synchronously** the instant "Yes" is pressed — *before* the scene swaps. That flips `GameState.hp <= 0` back to false, so the guard stops firing while the still-live field `Player` (state still `DOWN`) lives on for the transition frames. `take_damage()` is the universal damage choke point (enemy melee, projectiles, techs all funnel through it), so those transition-frame hits register as real damage.

## The fix
Decouple immunity from HP with a per-instance `_is_defeated` flag:
- Latched `true` on the death branch of `take_damage()` (the branch that drives HP to 0 and emits `died`).
- Checked as a hard no-op at the **very top** of `take_damage()` (augments, does not replace, the existing HP-coupled guard).

Because it isn't tied to HP, it survives the full-HP revive and stays immune through the whole defeat window (modal up, Yes pressed, scene transition). Clearing is implicit: `DefeatScreen.confirm_return()` does a full scene swap to the city counter — the field Player is freed and the city spawns a fresh Player (`_is_defeated` defaults false); the city has no combat. A defensive `_is_defeated = false` also lives on the `_respawn()` fall-out-of-world in-place revive path.

Orthogonal to the DefeatScreen modal_stack and the #413 meseta transaction — neither is touched.

## Spec
`spec/src/pages/states/player-death.astro` gains a normative **RFC-2119 damage-immunity clause**: from the moment of defeat (HP 0 / `died` emitted) through the return-to-city transition completing, the player MUST be damage-immune — `take_damage()` MUST be a no-op (no HP/PP change, no re-`died`), and immunity MUST NOT be coupled to HP (it must survive the full-HP revive). `_is_defeated` + the `take_damage` gate are added to "Implemented by".

## Tests
- **Unit (this PR):** `test_player_defeat_invulnerable()` in `test_runner.gd` mirrors the dodge-i-frame template — off-tree `PlayerScript.new()`, sets `_is_defeated`, asserts `GameState.hp` unchanged and zero `died` re-fires. Crucially it reproduces the ordering bug directly: revive to full HP with `_is_defeated` still true → `take_damage(25)` is still a no-op (the exact regression). Full suite green: `RESULTS: 2314 passed, 0 failed`, zero warnings-as-errors (#291 gate clean).
- **Autopilot (coordinator):** the two-layer `PSZ_AUTOPILOT_DEFEAT` / `record_defeat.sh` probe — extended to fire a hit during the post-Yes transition — will be run by the coordinator; not run here.

### DIFF
--- a/scripts/3d/player/player.gd
+++ b/scripts/3d/player/player.gd
@@ -164,6 +164,16 @@ var current_state: PlayerState = PlayerState.IDLE
 var player_rotation: float = 0.0
 var walk_timer: float = 0.0
 
+# Defeat immunity (#469, spec /states/player-death). Set true the moment the
+# player is defeated (HP 0 / died emitted); makes take_damage() a hard no-op
+# for the whole defeat window — modal up, "Yes" pressed, and the scene
+# transition back to the city. Decoupled from HP on purpose: the defeat
+# transaction revives to full HP synchronously on Yes-press (SessionManager
+# .defeat_return_to_city), so an HP-coupled guard stops firing while the
+# field Player still lives on for the transition frames. Cleared implicitly
+# by the city spawning a fresh Player (this flag defaults false).
+var _is_defeated: bool = false
+
 # Combo system — three-tier timing (#155, spec /mechanics/combos). Windows
 # are FRACTIONS of the current swing's animation, from per-weapon
 # WeaponComboConfig data. A press in a window QUEUES the next step, which
@@ -1035,6 +1045,10 @@ func _respawn() -> void:
 	player_rotation = 0.0
 	if model:
 		model.rotation.y = 0.0
+	# Defensive: this fall-out-of-world respawn restores the player to a live
+	# IDLE state in place, so clear the defeat immunity too (#469) — a genuine
+	# revive-in-place path, unlike the defeat exit which swaps to a fresh Player.
+	_is_defeated = false
 	transition_to(PlayerState.IDLE)
 
 
@@ -2508,6 +2522,14 @@ func _on_animation_finished(_anim_name: String) -> void:
 
 # Public API for external systems
 func take_damage(damage: int, _knockback: Vector3 = Vector3.ZERO) -> void:
+	# Defeated — hard no-op for the whole defeat window (#469, spec
+	# /states/player-death). Decoupled from HP so it survives the full-HP
+	# revive the defeat transaction applies on "Yes"; the field Player stays
+	# immune through the return-to-city transition. Cleared by the city
+	# spawning a fresh Player.
+	if _is_defeated:
+		return
+
 	# Already dead — ignore further hits
 	if current_state == PlayerState.DOWN and GameState.hp <= 0:
 		return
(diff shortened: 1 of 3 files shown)


kion-dgl:
## Autopilot: defeat probe ✅

Ran the dedicated defeat probe on this branch (fresh boot → Search & Rescue → warp → valley_field → `PSZ_AUTOPILOT_DEFEAT`):

```
godot rc=0 — pass ([sanity] DONE ok)
```

The full defeat→respawn→return-to-city flow completes end-to-end with the `_is_defeated` gate in place — no regression to the defeat transaction (meseta penalty + full-HP revive + city arrival).

The *specific* no-damage-during-the-transition-window behavior is proven deterministically by the unit test `test_player_defeat_invulnerable()` (revive to full HP while `_is_defeated` stays true → `take_damage()` still a no-op). Two layers green.

PR #438, opened 2026-06-28
fix(companion): key follow animation to measured speed, not player state (#420)

kion-dgl:
## Root cause

The companion NPC played its **run** clip while standing still during a player attack. In `scripts/3d/elements/companion_npc.gd`, `_process_follow` chose the companion's locomotion clip from the *player's* delayed `PlayerState`, not the companion's own motion:

```gdscript
var is_moving: bool = delayed_state >= 1     # ~line 411
...
if delayed_state == 1: _play_companion_anim("walk")
else:                  _play_companion_anim("run")
```

`PlayerState` is `IDLE=0, WALKING=1, RUNNING=2, ATTACKING=3, …`. A rooted player attack sets `ATTACKING (3)`, which is `>= 1` (so the companion tried to "follow" a player who wasn't moving) **and** `!= 1` (so it fell into the `run` branch). The trail-follow position never advanced — the companion ran in place. The animation was keyed to the player's *intent* instead of the companion's measured *displacement*.

## The fix (why)

Decouple the **position** decision from the **animation** decision in `_process_follow`:

- **Position** MAY still consult the player state, but only true locomotion states advance the trail: `delayed_state == 1 or delayed_state == 2`. A rooted `ATTACKING`/`DAMAGED`/`DOWN`/… no longer drives a follow-move.
- **Animation** is now a pure function of the companion's *own* planar displacement. New `_select_locomotion_anim(prev_pos, curr_pos, delta)` returns `wait`/`walk`/`run` against `IDLE_EPS` (0.15 m/s) and `RUN_EPS` (4.0 m/s), excluding vertical-only motion (`Vector2(moved.x, moved.z)`). **Zero planar displacement always yields `wait`**, so a stationary companion can never play a locomotion clip regardless of player state. The existing `ANIM_HOLD_TIME` (0.30 s) debounce is preserved.

Side benefit: because the clip now tracks real motion, the resume-from-freeze blend ramps `walk → run` naturally instead of snapping straight to `run`.

## Spec

New RFC-2119 page **`/states/companion`** ("Companion Locomotion") documents the position-vs-animation split, the speed→clip table, the *MUST-NOT-run-while-stationary* invariant, and a mermaid diagram noting the player attack has no edge into walk/run. Linked from `/states` and cross-referenced to `/states/player-state` (the enum it deliberately does **not** read for animation). Does not contradict the existing player-state contract.

## Verification

[...]

Issue #420, opened 2026-06-28
Companion NPC plays run animation while stationary during player attacks


kion-dgl:
**Source:** [playtest report 2026-06-28T06-11-12](https://psz.onl/reports/2026-06-28T06-11-12-084Z-c9c52c48) — Rozalin, v0.37.15

## Observed
> "If you have an NPC with you, when you do attack animations, the NPC just stays in position but does the run animation. If the NPC isn't actually moving positions, don't suddenly do the run animation while technically being stationary. It looks unnatural."

When the player triggers an attack, the accompanying companion NPC plays its **run** animation despite not changing position.

## Proposed spec
The companion's locomotion animation MUST be driven by its **actual frame-to-frame displacement**, not by the player's state or its follow-target:

- The companion MUST play `idle` when its planar speed is below a small threshold (e.g. `< 0.15 m/s`), regardless of what the player is doing.
- It MUST play `walk`/`run` only while actually translating, with the existing `ANIM_HOLD_TIME` (0.3s) debounce to avoid flicker on micro-stops.
- A player attack MUST NOT, on its own, change the companion's locomotion animation.

## Likely code
`scripts/3d/elements/companion_npc.gd` — locomotion anim is selected around `_current_anim` / the `run`/`walk`/`idle` map (lines ~63–166). Check what gates the run state; it appears to react to player attack/target state rather than the companion's own velocity. Suspect the follow-target snaps when the player roots for an attack, briefly registering non-zero intended velocity even though `move_and_slide` yields ~0 displacement. Drive the anim off measured displacement (`global_position` delta / `velocity.length()` post-move), not intent.

## Tests (two-layer)
- **Unit (test_runner):** seed a companion at rest next to a stationary player, fire `player attack`, step N physics frames, assert `companion._current_anim == "idle"` throughout.
- **Autopilot probe:** during a combat sanity run, assert no companion run-anim is selected in a frame where the companion's displacement is ~0.

---

## Expected behavior (RFC-2119 spec contract)

The companion's locomotion animation is a pure function of its **own measured planar speed** for the frame, never of the player's action state.

| Measured planar speed (post-`move_and_slide`) | Animation the companion MUST play |
|---|---|
| `< IDLE_EPS` (≈ 0.15 m/s) for ≥ `ANIM_HOLD_TIME` (0.30 s) | `idle` |
| `IDLE_EPS … RUN_EPS` | `walk` |
| `> RUN_EPS` | `run` |

[...]

PR #552, opened 2026-07-23
Fix companion run/walk animation flap with a hysteresis band (#463)

### LINKED ISSUE #463: Companion alternates run/walk animations while moving (surfaced testing PR #438)
From Rozalin's playtest of PR #438 (2026-07-02):

The original #420 bug (run animation while stationary) is fixed by PR #438. But while the companion is moving, it now appears to alternate between run and walk animations, which looks wrong.

Reproduction is inconsistent — while recording, it did not reproduce reliably, so it may be a pre-existing issue that #438's testing merely surfaced rather than a regression from #438. She could not reproduce it on 0.38.31 (pre-PR) so far, which weakly points at the PR. Comparison recordings are in the playtest thread.

Next step: determine whether it reproduces on main without #438 (e.g. drive the companion at speeds near the run/walk threshold — the PR keys animation to measured speed, so hysteresis around that threshold is the likely mechanism). PR #438 stays unmerged until this is resolved.

kion-dgl:
Closes #463.

## Why

From Rozalin's playtest of #438: while the companion is moving it visibly alternates between the run and walk animations.

Root cause (traced, not guessed): `CompanionCombat.locomotion_clip` chose the clip from a **single hard speed threshold** (`RUN_EPS = 4.0`) applied to a **noisy per-frame measured planar displacement** (`companion_npc._select_locomotion_anim`). The companion's steady follow speed is ~6 m/s but the per-frame measurement dips through 4.0 constantly, so the selector's *output* toggled run↔walk every frame.

The existing 0.3s `ANIM_HOLD_TIME` debounce did **not** fix this — it only rate-limits the flap to ~3 Hz. The debounce gates how often the clip can change; it can't stop a change when the selector it feeds keeps flipping. The output still visibly toggles.

## Fix — speed hysteresis / deadband (selector stays pure)

Give the walk↔run split a hysteresis band:

- Replace `RUN_EPS` with **`RUN_ENTER = 4.0`** (switch walk→run only *above*) and **`RUN_EXIT = 3.0`** (switch run→walk only *below*).
- In the 3.0–4.0 band the pure selector returns the **current clip unchanged** (hold), so a measured speed hovering at the boundary can't flap the animation.
- `locomotion_clip(intent_moving, planar_speed, current_clip)` gains a `current_clip` arg; the caller threads in the companion's `_current_anim`.
- The `IDLE_EPS → "wait"` #420 veto is unconditional and still takes precedence over the run-hold. Coming out of `wait` uses the enter threshold, so a body drifting through the band picks `walk`, never a phantom `run`.

The 1 m/s band sits below the ~6 m/s steady follow speed. **Its width is the one value worth confirming on-device.**

The consts on `CompanionCombat` remain the single source, shared with the autopilot run-in-place tripwire (which keys only on `IDLE_EPS` — semantics unchanged).

## Spec

`/states/companion` gains a normative **"Walk↔run hysteresis (#463)"** clause (RFC-2119: MUST enter above `RUN_ENTER`, MUST exit below `RUN_EXIT`, MUST hold `current_clip` in the band), plus updated intent×speed table, mermaid edges, and Implemented-by signature/thresholds.

## Tests

Unit (the definitive layer — `test_companion_anim_from_measured_speed`, seeded/off-tree):

- Existing call sites updated for the new signature and kept passing.
- Threshold cases: enter run only above `RUN_ENTER`, hold in band, exit only below `RUN_EXIT`, `wait`→band picks walk.
- **Anti-regression assertion**: an oscillating band-straddling sequence (`3.9, 4.1, 3.9, 4.1, 3.5, 4.2, 3.1, 3.8, 3.9`) threaded back through `current_clip` each step must **stop changing the clip once settled in run** (zero transitions after settle) — this deterministically proves the flap is gone headlessly.
- Honest transitions preserved: clear climb 0→5 → run; clear drop 5→0 → walk → wait.

`RESULTS: 2323 passed, 0 failed`, zero `Warning treated as error`.

Autopilot layer: a walk↔run flap-rate tripwire on the `PSZ_AUTOPILOT_COMPANION_COMBAT` probe will be handled by the coordinator (no companion-bearing autopilot phase exists to run here yet).

### DIFF
--- a/scripts/3d/elements/companion_combat.gd
+++ b/scripts/3d/elements/companion_combat.gd
@@ -69,6 +69,20 @@ static func locomotion_clip(intent_moving: bool, planar_speed: float, current_cl
 	return "run" if planar_speed > RUN_ENTER else "walk"
 
 
+## Commanded FOLLOW speed (spec /states/companion): the companion holds `ring`
+## metres behind the player, and its speed is a CONTINUOUS ramp of how far it sits
+## OUTSIDE that ring — 0 at/inside the ring, climbing linearly over `ramp` metres to
+## `cap`. Continuity is the whole point: fed through locomotion_clip, this eases the
+## gait wait→walk→run and back with no dead zone and no flap (the run/walk swing the
+## old measured-displacement path produced when the companion caught up and stalled).
+## `cap` is set just above player run so the trailing gap holds instead of growing.
+## Pure — test_runner pins the ramp off-tree.
+static func follow_speed(distance: float, ring: float, ramp: float, cap: float) -> float:
+	if ramp <= 0.0:
+		return cap if distance > ring else 0.0
+	return cap * clampf((distance - ring) / ramp, 0.0, 1.0)
+
+
 ## True when pos is within LEASH_RADIUS (XZ) of the player.
 static func within_leash(pos: Vector3, player_pos: Vector3) -> bool:
 	return Vector2(pos.x - player_pos.x, pos.z - player_pos.z).length() <= LEASH_RADIUS
--- a/scripts/3d/elements/companion_combat.gd
+++ b/scripts/3d/elements/companion_combat.gd
@@ -30,9 +30,14 @@ const COMPANION_WEAPON_TYPES: Dictionary = {
 
 ## Locomotion clip thresholds (m/s of the companion's OWN measured planar
 ## speed). The pure selector below and companion_npc's tripwire share them so
-## there is a single source (spec /states/companion).
-const IDLE_EPS: float = 0.15  # below this the companion plays "wait"
-const RUN_EPS: float = 4.0    # above this the companion plays "run"
+## there is a single source (spec /states/companion). RUN_ENTER/RUN_EXIT form a
+## hysteresis band: measured speed is a noisy per-frame displacement, so a single
+## run/walk threshold flaps the clip when it hovers there (#463). The 1 m/s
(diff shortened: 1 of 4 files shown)


kion-dgl:
## Autopilot ✅ (regression) — companion runs clean with the hysteresis change

The **definitive** anti-flap proof is the unit test (oscillating band-straddling sequence → zero clip transitions once settled). For the autopilot layer:

- **Plain SR sanity → pass** (`godot rc=0`, `[sanity] DONE ok`, 99 checkpoints): companion `kai` spawns, follows, and animates end-to-end through Search & Rescue — no regression from the `locomotion_clip` signature change.

**Note (pre-existing, unrelated):** the dedicated `PSZ_AUTOPILOT_COMPANION_COMBAT` probe currently fails with `FAIL: companion-combat probe — wolf missing from EnemyRegistry` — it wants a wolf as an attack target in the SR first field cell, which isn't registered there. That's a probe-setup gap independent of this PR (this change touches only `locomotion_clip` + its caller + spec + test — nothing in `EnemyRegistry`/enemy spawning; the companion itself spawned and loaded locomotion anims fine in that run). That probe also only asserts a hit lands, not animation flapping, so it isn't the right oracle for #463 regardless. Worth a separate follow-up to fix the probe's target setup.

**Reminder:** the 1 m/s deadband (RUN_EXIT 3.0 / RUN_ENTER 4.0) is the one tuning value worth eyeballing on-device.

PR #567, opened 2026-08-02
Reyburn: gate flight to low HP (less boss jank, psz-re B2)

kion-dgl:
## Why
`reyburn_boss.gd` triggered its flight/land-slam cycle every 22s from full health — invented behaviour. psz-re's **B2** finding measured Reyburn **grounded from 100%→86% HP** with no takeoff observed, and the real threshold is unconfirmed. A dragon flying off mid-fight at full HP is a likely source of the "jank" noted on the boss.

## Change
Gate flight to below half HP (`FLIGHT_HP_FRAC = 0.5`); the flight timer still paces takeoffs *within* that low-HP phase. `reyburn_boss.gd` only — no shared code, no other boss affected.

## Verification
- Parse + code-health clean. Valley free-roam autopilot still clears A→E→B→Reyburn to `DONE ok` (boss spawns, is killed, field completes).
- The flight *behaviour* change itself is only observable in a real fight — worth a on-device look. The exact takeoff threshold is a follow-up once psz-re traces it (B2).


kion-dgl:
## Autopilot regression — Reyburn / free-roam

Change is isolated to `reyburn_boss.gd`, exercised by the Valley boss room. Ran the Valley free-roam phase (A→E→B→Reyburn):

- **gurhacia (Valley): ✅ PASS — `DONE ok`** — Reyburn spawns, is killed, field completes; no regression from the flight gating.
- code-health + GDScript parse clean.

Quest phases are unaffected (reyburn_boss is only used by the free-roam boss room + the debug boss quest). The flight *behaviour* (only takes off below half HP now) is only observable in a live fight, so it's the one thing worth an on-device look — the autopilot instakills the boss so it can't exercise the takeoff.

PR #568, opened 2026-08-02
Free-roam: boxes + traps at authored positions (psz-re Q6)

kion-dgl:
## What

Free-roam fields placed **boxes at blind ring positions** and had **no traps at all**. psz-re's Q6 decoded the original game's per-room object table (`set/<NN>/<v>/<room>/…_s.rel`) — 24-byte records carrying a room-local `(x,y,z)` in the same frame as `room_doorways.json`. This PR drops the **real authored positions** in.

The spawner even carried a standing TODO for exactly this ("*until the measured object table … lands and we can drop in the exact authored positions*").

## Result

- **486 boxes** now sit where the original game puts them (was blind ring placement).
- **435 traps** exist for the first time — all 5 psz-re types (needler / burn / gun / elemental / capture).

## How

| File | Change |
|---|---|
| `data/re_reference/room_objects.json` | Authored objects extracted from the psz-re set table (267 rooms). Groups 0–4 + ungrouped files are verbatim; the runtime-randomised **group 5** is left out. |
| `scripts/tools/refield/apply_authored_objects.py` | Bakes authored box/trap positions into every field JSON (`--check` drift guard). Enemies untouched; mesh is never rotated so room-local → world directly. |
| `needle_trap.gd` | Parameterised with `damage`/`trap_kind`/`element` (data variation, not new leaf classes). Four contact traps share it (needler 25, burn 50 from Q6); capture → bear trap. |
| `cell_object_spawner.gd` | Wires damage/element through, floor-validates traps, spawns traps **disarmed under autopilot**. Also fixes a latent box save/restore bug (matched the floor-snapped `y`; now x/z only). |
| `record_free_roam.sh` | `NO_BOXES` for the nav harness — authored boxes are solid breakable content on the walk backbone the autopilot can't attack. Traps are Area3D-only + disarmed, so they don't block. |
| `spec/states/free-field` | Documents authored placement + the trap-type map. |
| `test_runner` | `test_freeroam_authored_objects`: every authored box is baked into its cell, traps carry damage/kind, and `needle_trap` honours its damage param. |

## Verification

- Test suite **2387 passed / 0 failed**.
- gurhacia freeroam autopilot **DONE ok** with real objects (15 traps disarmed-spawned, no crash).
- `validate_quests` (35 files) + baseline-debt + `check_asset_refs` + `check_repo_layout` all clean.

## Deliberately deferred

- **Group-5** objects (runtime-randomised pool) — skipped rather than reproducing the roll.
- Burn/gun/elemental reuse the needle-trap model until dedicated assets exist.
- Keys/warps/containers/switches — keys & warps are handled by the field's own gate/section systems; containers/switches left for a follow-up.


kion-dgl:
## 🟢 Autopilot regression matrix — PASS (16/16 + freeroam)

Ran the full matrix on this branch (local pack, headless). Every phase green:

| Phase | Status | Phase | Status |
|---|---|---|---|
| boot | ✅ | fo | ✅ |
| first-mission (SR) | ✅ | sis / static | ✅ |
| pp_canon | ✅ | it (investigate tower) | ✅ |
| pp_backtrack | ✅ | heretic | ✅ |
| as_canon | ✅ | csy (control system) | ✅ |
| as_moon | ✅ | tbs (broken seal) | ✅ |
| as_sol | ✅ | dc (dark castle) | ✅ |
| doe | ✅ | shops | ✅ |

**freeroam (gurhacia)**: ✅ `DONE ok` — clears A→E→B→Reyburn **with the new authored objects** (15 traps disarmed-spawned, boxes skipped by the nav harness via `NO_BOXES`).

**Result: 16/16 quest phases pass, freeroam pass, 0 failures.** The authored box/trap placement, the parameterized `needle_trap`, and the box save/restore x/z fix did not regress the quest chain (guild quests are untouched — this PR only rewrites `data/field_quests/*`).

Not covered by the gating matrix: ozette/paru freeroam, which fail on a **pre-existing waypoint-authoring gap** (`s02a_tc3`, `s05a_lb1`) unrelated to this PR — addressed in a follow-up PR.

kion-dgl:
Superseded by #591, which merged as 93257b66.

#591 does what this does from a larger extract and, critically, implements the part this one skipped: **group 5**. This PR's description records "Group-5 objects (runtime-randomised pool) — skipped rather than reproducing the roll", and group 5 is where the traps live — 591 of the 1,122 trap records in set `d` carry `g: 5`, and psz-re reads the group as 92% factory index 19 (the elemental trap). A port that builds groups 0–4 and stops places almost no traps, which is why #591's roll — `rand(100)` against `[40,20,20,20]`, Fisher-Yates, take that many, 40% of rooms taking nothing — was the load-bearing half rather than the skippable one.

The other differences, for the record: 290 rooms vs 267, and #591 leaves `data/field_quests/*.json` untouched (it authors the *generated* path) rather than baking positions into the static quest files, so the two had six conflicting field JSONs between them.

Nothing here is lost. The needle-trap parameterisation, the box save/restore x/z fix and the disarmed-under-autopilot spawning all landed in #591 in equivalent form, and the `NO_BOXES` nav-harness flag is now the pattern #593 extends to walls. Thanks for the regression matrix run on this branch — 16/16 plus freeroam was useful evidence for the approach even though this particular diff isn't the one that shipped.

Issue #579, opened 2026-08-08
psz-re: traps — which types exist per field, and how many per room?


kion-dgl:
**From psz-godot.** We added procedurally-placed floor traps to the generated free fields, playtested them, and pulled them straight back out: there were far too many, and every one was a needle trap or a bear trap because those are the only two trap elements the game code has. The placement was invented — a hardcoded two-entry list, 0–2 per combat room, seeded — with no data behind it. It is gone now and will stay gone until this issue has an answer.

## What we would need to put them back

1. **The trap pool.** Which trap types can a field room actually contain? PSZ ships more trap art than we place — `o0c_needle` (needle), `o0c_torabasami` (bear), `o0c_poisonm` (poison), `o0c_gun01` / `o0c_gun02` + `o0c_bullet01` (turret + projectile), `o0c_clay1` (a ~21-unit trap volume, role unclear), `o0c_bombcont` (explosive container). Is there a per-field or per-room-code trap table the way `enemy_room_assignment` works for waves?

2. **Density.** How many traps does a room get, and is it a count, a probability per slot, or a budget shared with boxes/enemies? `level_generation_algorithm.json` notes a block of percentage-looking values that are "the best candidate yet for the box / enemy / trap placement weights" — if those decode, this is answered.

3. **Placement.** Are trap positions authored per room model (like the boss-arena `enemy/deploy/*.rel` XYZ triples), or generated? Ours are on a ring around the room centre, which is a guess.

4. **`o0c_clay1` specifically** — it is textured from the shared `o0s_*_trapc` sheet rather than any field palette, and it is wall-sized rather than prop-sized. Is it a trap at all, or scenery that happens to sit in the trap texture set?

## Not blocking

Nothing depends on this — the fields play fine without traps, and we would rather ship none than ship the wrong density again. Answer whenever it surfaces naturally in the backlog.

Related: #578 (boxes), #562.

kion-dgl:
**From psz-re.** Answered. Commit `23ddc66` (psz-re) — `docs/godot-field-parity.md` §8, `data/trap_placement_per_room.json`, `tools/traps.py`, three gated checks in `tools/checks.py`.

Short version: **you were not wrong to pull them out, but the reason is not density — it is that the two trap elements you had are the wrong two, and the placement is authored.** There are five trap types, one of which is four traps in a trenchcoat, and **the group that #578 told you to leave out is the trap group.**

---

## 1. The trap pool — five object types, and the game names them itself

[...]

Issue #578, opened 2026-08-08
psz-re: object placement per room — how many boxes, and where?


kion-dgl:
**From psz-godot.** Free fields are now generated at runtime (rooms, gates, keys, enemy waves all come from decoded data), and object placement is the one part still made up. Anything you can pin down here goes straight into the generator.

## What we already use, and where it came from

Enemy waves are solid — `enemy_room_assignment.json` gives per-room-code templates + weights for every field area, `enemy_wave_templates.json` gives each template's enemy list, and the RE internal name resolves to a psz-godot enemy through the `model_id` on each `data/enemies/*.tres` (65 of the 69 table entries land). No questions there.

Boxes are the gap. `data/re_reference/level_generation_algorithm.json` already records:

> `FUN_02082814(gen, internal +0x03)` — a lookup, **NOT a box count**. The refuted "boxes = 2 × byte at +0x31" guess was reading a table index.

and, about a nearby block of values:

> unidentified, but they read as percentages and are the best candidate yet for the box / enemy / trap placement weights kion asked about

So today the generator places a flat **2 boxes per combat room**, everywhere, and that number is a placeholder with nothing behind it. The only real data we have is a hand-transcribed observation of one Valley area-A layout (`AREA_A_BOXES` in `scripts/tools/refield/build_valley_field.py`: B3=4, B2=0, B1=2, A2=3, C4=4, C1=4), which does not transfer to a randomly generated layout.

## Questions

1. **Those percentage-looking values** — are they in fact the box/enemy/trap placement weights? If so, what is the table, and what are they weighting (count per room? probability per candidate slot? per room shape?).
2. **Is box count a function of the room**, the field, the difficulty, or the run seed? The Valley area-A observation has a 0 in it (B2), so whatever drives it can produce an empty room.
3. **What does the `+0x03` lookup index into?** Knowing the table it selects from may answer 1 and 2 together.
4. **Positions.** We place objects on a ring around the room centre and floor-snap them, which is entirely ours. Is there a per-room-model object spawn table (something like the `enemy/deploy/*/s0Xz_na1.rel` XYZ triples, but for the a/b rooms)?

## Why it matters

We shipped procedurally-scattered traps on this same basis and they were immediately, obviously wrong in play — far too many. They have been removed. We would rather leave boxes at a labelled placeholder than repeat that, so no answer is a fine answer; a partial one ("it is per room shape, table at X") is still useful.

Related: #562.

kion-dgl:
[...]

Issue #575, opened 2026-08-08
Player-placed traps: the four ball models, and why they are not the field trap meshes


kion-dgl:
## The models

The traps a CAST places are **`o0c_burst01`..`o0c_burst04`**, now viewable at:

https://dashgl.github.io/psz-asset-viewer/effects/item_effects/

They were hard to find because they are **not** in the field object sets — they live in the item/effect archive, and the word "trap" appears nowhere in a filename. It only shows up in the *texture* names, `o0s_1_trapb01..04`, where `trapb` = **trap ball**.

## What they actually are

All four are **one shared mesh with four skins**. The four model files are byte-identical except six bytes, and every one of those six is an ASCII digit inside a name. So there is a single ball, recoloured:

| model | measured texture colour |
|---|---|
| `o0c_burst01` | green |
| `o0c_burst02` | orange |
| `o0c_burst03` | magenta |
| `o0c_burst04` | blue |

**Which ball is which trap is NOT established.** There are exactly four trap items — Heat, Ice, Light, Heal — and both the item and object name lists run in that order, so `01→Heat` is the tempting reading. The colours contradict it: orange reads as Heat and blue as Ice, which would make the order something like Heal, Heat, Light, Ice. Treat the mapping as open. A savestate with a known trap on the ground settles it in one shot.

## Do not confuse these with the field traps

The trap meshes already in storybook — `o0c_torabasami` (a とらばさみ leg-hold snare), `o0c_needle`, `o0c_gun01`/`gun02`, `o0c_poisonm` — are **environmental** traps, pre-placed in rooms. They are a different family:

- **player traps** — 4 items (Heat, Ice, Light, Heal) → the ball models above
- **level-placed player traps** — the same four, pre-placed as room objects
- **environmental traps** — Capture, Gun, Burn, Needler; object-only, no item form, and these are the mesh models

The game's object vocabulary names eight traps; the item list names only four. That subset relationship *is* the taxonomy.

## Material setup (matters for parity)

The exported materials are **unlit**, `alphaMode: MASK`, **double-sided**, nearest-neighbour filtered, and use **MIRRORED_REPEAT on both axes**. The mirroring is not specific to these — see the separate texture-wrap issue; it affects most imported models.

## Acceptance

- [ ] Four trap balls placeable, one shared mesh + four materials
- [ ] Unlit / alpha-mask / double-sided / nearest filtering preserved
- [ ] Mirrored wrap correct on both axes
[...]

PR #591, opened 2026-08-12
Field objects + traps at authored positions, and the elemental trap (psz-re #578/#579/#575)

kion-dgl:
## Why

Free-roam fields put **boxes on a ring** and placed **no traps at all**. Both were approximations of a placement pass the original game does not have.

psz-re decoded `set/<NN>/<v>/<room>/*.rel`, the per-room object table: every box, wall and trap in the game is **authored** there with a room-local `(x, y, z)` in the same frame as `room_doorways.json` plus a 16-bit facing. There was never a box-count formula to find — the one previous reading of it was refuted because it was a *layout index*, not a count.

## What selection survives, and why it's the load-bearing part

A set file is **six groups**:

| | |
|---|---|
| groups 0–4 | built **verbatim**, whichever a layout mask names. The layout index is a `rand(100)` draw against a depth-banded weight row, with an eligibility rule. |
| **group 5** | **rolled** — `rand(100)` against `[40,20,20,20]` for a count of 0–3, shuffle, take that many. So **40% of rooms roll nothing.** |

**Group 5 is the trap group** — 99.5% of the original's trap pool sits in it. An implementation that built 0–4 and stopped would place almost no traps, so this is the part that had to be right rather than the part that could be skipped. Every draw goes through the generator's seeded RNG, so a seed still reproduces a field exactly.

## The elemental trap is the player's trap item

psz-re decoded the original's class this week: the four trap items a CAST drops (**#575**) and the four elemental traps authored into every room (**#579**) are **one object** — factory index 19, type `0x1C13` — separated by a single byte in the object's parameter block. So this adds `TrapBall.field_placed` rather than a second script, and the field form behaves as the ROM says:

- **invisible while dormant** — the original's factory table gives it no model at all
- **arms** when a player comes within 4.0 units, **rises 2.5 units**, becomes visible
- **fuses 45 / 30 / 15 frames** on normal / hard / v_hard — the original gets *faster*, not stronger, with difficulty — then detonates on whoever is still inside
- **heal restores 50% of max HP**, now measured rather than the guess it was

A **CAST** sees a dormant trap. Anyone else sees one under **Trap Vision**, which is now a usable consumable; a second one extends rather than stacks and cannot cut a longer one short.

The other four trap types share `needle_trap` by data (Needler 25, Burn 50 — both measured constants; capture keeps the bear trap) rather than four leaf classes.

## Files

| File | Change |
|---|---|
| `data/re_reference/room_objects.json` | **new** — 290 rooms, 2,726 authored objects. Every cell of every field quest resolves; no room code falls back to the ring today. |
| `scripts/tools/refield/import_re_objects.py` | **new** — the importer, with a `--check` drift guard so the file cannot go stale against psz-re silently. |
| `scripts/3d/field/field_population.gd` | authored placement + the layout/group-5 roll, replacing the ring. Ring kept only as a fallback for a room code invented later. |
| `scripts/3d/elements/trap_ball.gd` | `field_placed`, the measured fuse table, the rise, and the CAST / Trap Vision reveal. |
| `scripts/3d/field/cell_object_spawner.gd` | spawns all 8 authored trap kinds; elemental traps spawn inert under autopilot. |
| `scripts/3d/elements/needle_trap.gd` | `damage` / `trap_kind` params (data variation, not new leaf classes). |
| `scripts/autoloads/inventory.gd` | Trap Vision is usable. |
| `scripts/tools/check_generated_objects.py` | **new** — the second test layer, below. |
| `spec/src/pages/states/free-field.astro` | the normative contract for authored placement, the roll, and the trap families. |
| `web/public/field-dumps/generated-fields.json` | regenerated — this is where the change is visible: boxes and traps now appear in generated fields. |

## Verification

- **Test suite: 2559 passed, 0 failed.**
- New seeded tests cover the table's coverage, the roll's seed-determinism, the caps, the 0–3 group-5 distribution (44% rolled nothing over 400 trials), the fuse table at every difficulty, and the reveal rule.
- **Free-roam autopilot `gurhacia`: `DONE ok`.** Honest caveat: that path loads the *static* `field_quests/*.json`, so it is a **regression** check, not coverage of the new code.
- Which is why `check_generated_objects.py` exists — unit tests call `authored_objects()` directly and cannot catch the *generator* failing to call it. It reads the regenerated dump of the real generator: **392 cells over 7 areas, 804 containers, 293 traps over all 8 families**, every authored position inside its room.
- `code_graph --check` (no new dup/dead/complexity/coupling), `check_baseline_debt`, `validate_quests` (35 files), `check_repo_layout`, `check_asset_refs`, `import_re_objects --check`, spec build — all clean.

**That last check earned its bound.** Held at ±22 it failed on 34 of 2,726 objects; the furthest is 24.29 and **zero** sit past 25 — they are standing in the doorway mouths, which psz-re independently measures as stubs running 22 → 25. An agreement with the door geometry from a file that knows nothing about it.

## Deliberately not done

- **Authored walls are imported and not placed.** As measured as the boxes; the reason is gameplay — the autopilot's walk backbone cannot attack a blocker, and a wall across the line between two doorways is a soft-lock. Gated behind `FieldPopulation.PLACE_WALLS`.
- **What arms a field trap is not decoded upstream.** "A player walks close" is the obvious reading of a thing that rises out of the ground on a fuse; the 4.0 units is the one radius psz-re publishes (the heal element's player scan), reused. Both are in the spec's approximation list, not presented as measured.
- **Field-trap damage is ours** — the original carries no per-instance trap parameters at all, so there is no number to port.
- **Static `field_quests/*.json` are untouched.** They are built by `refield/build_*.py` and still carry their own object lists; only the generated path is authored so far.
- Boss (`z`) and transition (`e`) rooms keep enemies-only population.

## Overlaps #568

**#568** ("Free-roam: boxes + traps at authored positions") does part of this from a smaller extract — 267 rooms, 486 boxes, and **group 5 skipped**, which is where the traps are. This is built on current `main` and does not include #568's changes, so the two conflict: pick one. Happy to rebase this onto #568 instead if you'd rather keep its baked-into-field-quests approach.


kion-dgl:
## Layout index 4 is unreachable, so group 4 never spawns

Reading the shipped table against the selection rule: **layout index 4 cannot be
drawn by either path**, and it is the only mask that names group 4.

**The two paths, both capped at 3:**

- `_weighted_index()` walks a weight row and returns an index into it. Every row
  in `layout_weights_by_depth` is **4 entries** (`lt4 [50,20,20,10]`,
  `4_6 [40,20,20,20]`, `ge7 [25,25,25,25]`), so the draw returns 0..3.
- The ineligible-mask fallback in `_pick_layout()` is explicitly
  `if i <= 3 and i > best`, so it can't return 4 either.

**Why that costs something.** `layout_masks` is `[0x21, 0x23, 0x25, 0x28, 0x30]`:

| index | mask | groups |
|---|---|---|
| 0 | 0x21 | 0, 5 |
| 1 | 0x23 | 0, 1, 5 |
| 2 | 0x25 | 0, 2, 5 |
| 3 | 0x28 | 3, 5 |
| **4** | **0x30** | **4, 5** |

Group 4 appears in mask 0x30 and nowhere else. In `room_objects.json` group 4
holds **146 of the 2,019 grouped objects (7.2%)**, and one room —
`s07a_ga1_d` — is *entirely* group 4, so it can never place anything and falls
through to the `BOXES_PER_ROOM` ring every time.

**The question for psz-re**, since I can't tell from this side which reading is
right:

1. Is the weight row at `0x020F2E70` really **4** entries wide, or 5 with the
   import taking the first four? A 5-wide row with the same shape would make
   index 4 drawable and light up group 4.
2. Is the `<= 3` bound in the eligibility fallback a bound the game actually
   carries, or was it inferred from the 4-wide row? If both are genuinely 4-wide
   in the ROM, then index 4 and group 4 are dead in the *original* too — which
   is a much stronger claim (the level files would be carrying 146 objects the
   game never builds) and worth its own confirmation before we mirror it.

Either answer is fine for this PR — if 4-wide is correct we just want a comment
in `room_objects.json`'s `the_rule` saying index 4 is unreachable **by design**,
so the next person reading the mask table doesn't re-derive this. What we don't
want is the current state, where the file ships a fifth mask and a group nothing
can select and says nothing about it.

Issue #594, opened 2026-08-13
Import and place the rest of the authored objects: fences, switches, keys, mspack


kion-dgl:
**Blocked upstream on psz-re — but on serialization, not on decoding.** A task has been queued in the psz-re session to emit these; this issue is the psz-godot half, ready to start the moment the dump lands.

## What is missing

#591 imports boxes, rare boxes, walls and the eight traps from psz-re's per-room set-table decode. That is not all the table holds. psz-re's `data/object_factory_tables.json` → `placement_census_by_model`, over the whole corpus (all sets, 1,491 rooms):

| Model | Records | Imported today |
|---|---|---|
| `o0c_key` | 3,972 | ❌ **keys** |
| `o0c_trebox` | 3,115 | ❌ (psz-re: "what it is, is open") |
| `o0c_shfence` + `o0c_fence` | 1,211 | ❌ **fences** |
| `o0c_remswitch` + `o0c_switchf` + `o0c_switchs` | 705 | ❌ **switches** |
| `o0c_mspack` | 324 | ❌ |
| `o0c_clay1` | 248 | ❌ |
| 12 unresolved factory indices | 1,435 | ❌ (being decoded upstream) |

Measured from this side: across the 214 grouped rooms in set `d`, `group_sizes` counts **3,556** records and the two dumps classify **2,019** — so **43% of authored records never reach the game**, and 35 rooms import as completely empty and fall back to the `BOXES_PER_ROOM` ring (e.g. `s00a_nr1_d` declares 5 records in group 0 and we import none).

## Upstream ask (queued)

psz-re extends `data/object_placement_per_room.json` (or a sibling dump) to emit every classified kind on the same schema it already uses — `group, kind, x, y, z, angle` — so the importer can read new kinds by name the way it already reads the eight trap ids.

## The psz-godot half

1. **Importer** — `scripts/tools/refield/import_re_objects.py`: extend the kind map beyond `box` / `rare_box` / `wall` + `TRAP_IDS`. Keep `--check` honest so the file cannot go stale.
2. **Spawner** — `scripts/3d/field/cell_object_spawner.gd` already handles `fence` and `step_switch` for the static quests; wire the imported kinds onto those cases rather than adding new ones. `o0c_mspack` looks like the meseta pack the spawner's `"meseta"` case already covers.
[...]

PR #597, opened 2026-08-13
Place the authored walls, gate them at spawn time (#593); document the unreachable layout 4 (#592)

### LINKED ISSUE #592: #591 follow-ups: restore trap-ball mirroring, document unreachable layout index 4
Two small follow-ups to #591. Both are in that PR's own files; if #591 has not merged yet they belong on the branch, otherwise they are a one-commit PR of their own.

## 1. Restore the trap-ball texture mirroring (regression)

`bfe33d87` ("Make the palette actually reach the traps, and mirror their textures", 2026-08-11) added `_apply_mirror_wrap()` + the `MIRROR_SHADER` preload to `scripts/3d/elements/trap_ball.gd`, specifically to fix the burst GLBs importing as CLAMP and rendering as the "grey angular mass with a coloured streak" #575 flagged.

#591's single commit sits directly on top of that and deletes it. On `main` the word "mirror" appears 15 times in `trap_ball.gd`; on the PR branch it appears **0** times, and `_load_ball()` now instantiates the model with no material work at all. `scripts/3d/shaders/mirror_repeat_effect.gdshader` is left with **zero consumers** as a result — `trap_ball.gd` was its only one — which is the tell that this was a rebase resolution taking the wrong side rather than a decision. `orphan_files --check` did not catch it.

**Fix:** restore `_apply_mirror_wrap()` and the `MIRROR_SHADER` const, and call it at the end of `_load_ball()` as `main` does.

**While you're in there:** #591 also walks `REST_HEIGHT` 1.85 → 1.6, undoing the "hang at the crown of the head" tune from that same day-old commit. That one *is* argued in a comment (the ball mesh is ~0.5 across so its centre sits under the crown), so it may be deliberate — worth confirming rather than reverting blind.

**Acceptance:** a placed trap ball renders with its mirrored texture again; `grep -c mirror scripts/3d/elements/trap_ball.gd` is non-zero; `mirror_repeat_effect.gdshader` has a consumer.

## 2. Note that layout index 4 is unreachable, in the data file

`data/re_reference/room_objects.json` ships five `layout_masks` (`[33, 35, 37, 40, 48]`) but the layout draw can only ever return 0–3, so mask `0x30` — the only mask naming group 4 — never gets selected. 146 objects in set `d` sit in group 4 and can never spawn; `s07a_ga1_d`'s contents are entirely group 4, so it always falls through to the `BOXES_PER_ROOM` ring.

**This is faithful, not a bug.** Checked against psz-re (see the discussion on #591): `data/level_generation_algorithm.json` reads room-record byte `+0x31` as `FUN_02082814(gen, depth)`, "a weighted random draw returning a category 0..3", table `0x020f2e70` = "twelve ints as three rows of four; every row sums to exactly 100", validated with "all 44 values across five fields are in 0..3". The `<= 3` bound in `_pick_layout()` is real.

What *is* open is upstream: psz-re's `object_placement_per_room.json` → `NOT_ESTABLISHED` records "the mission-guarded path that forces layout 4 is not modelled; free fields in the sample never took it". So group 4 belongs to a mission-guarded path, and free-roam generation correctly never sees it.

**Fix:** add a line to the `the_rule` array in `room_objects.json` saying layout index 4 is unreachable by the depth-banded draw by design, and that group 4 is reserved for the unmodelled mission-guarded path — so the next reader of the mask table does not re-derive this. Mirror it in the `authored_objects()` doc comment in `scripts/3d/field/field_population.gd`.

**Acceptance:** a reader of either file learns why there are five masks and four weights without leaving the file.

### LINKED ISSUE #593: Place authored walls: move the gate from generation to spawn time, wire PSZ_AUTOPILOT_NO_WALLS
Depends on #591 (or its successor) landing first — it is what imports the walls.

## Why

`data/re_reference/room_objects.json` already carries **196 authored walls** for set `d`, with their facing, imported from psz-re's set-table decode exactly like the boxes. They are not placed: `scripts/3d/field/field_population.gd` gates them behind `const PLACE_WALLS := false`.

The reason for the gate is sound — the free-roam autopilot's walk backbone cannot attack a blocker, and an authored wall across the line between two doorways is a soft-lock rather than a cosmetic complaint. But it means nobody can playtest against the real room layouts, and walls are a big part of what makes an authored room read as authored.

## The problem with where the gate sits

`PLACE_WALLS` is checked inside `_to_object()` — that is **generation** time, so the wall never enters the cell's object list at all. Two consequences:

1. The generated field *data* differs between an autopilot run and a player run. `web/public/field-dumps/generated-fields.json` and `scripts/tools/check_generated_objects.py` see a different world than the player does.
2. Cell contents persist. A cell first entered during an autopilot run is saved without its walls and restores that way forever.

## What to do

Move the gate from generation to **spawn** time, matching the pattern boxes already use — `scripts/3d/field/cell_object_spawner.gd:710`, `OS.has_environment("PSZ_AUTOPILOT_NO_BOXES")`:

1. Drop the `PLACE_WALLS` filter from `_to_object()` so walls flow into the object list like every other kind. Keep the const only if something else wants it.
2. Add the skip to the `"wall"` case in **both** `_spawn_fresh_cell_objects` and `_restore_cell_objects`, under `PSZ_AUTOPILOT_NO_WALLS` (or fold it into the existing `PSZ_AUTOPILOT_NO_OBSTACLES`, which is already semantically "keep collision off the walk backbone" — pick one and say which in the PR).
3. Set the var in `scripts/tools/autoplay/run_regression_matrix.sh`, `run_quest_matrix.sh`, `record_free_roam.sh` and `record_the_paru_pact.sh`, next to the existing `NO_BOXES` / `NO_OBSTACLES` lines.

## Open question to settle before playtesting

`_to_object()` emits no `destructible` key, and the spawner's `"wall"` case defaults to `true`. Whether the original's authored walls are breakable is not established — if they are not, a wall across a doorway line is exactly the soft-lock the current gate avoids, and this needs a placement guard (e.g. reject a wall whose position sits on the segment between two of the room's doorways) rather than just a toggle. Worth one look at psz-re's `wall_placement_per_room.json` before implementing.

## Verification

Two layers, per the working agreement:

- Seeded unit test: a generated cell's object list **contains** its authored walls (the data no longer depends on the env var), and the spawner skips them when the var is set.
- Autopilot: full regression matrix green with `NO_WALLS` set, **and** one free-roam run with it unset that either clears or fails loudly — that run is the evidence for whether walls soft-lock the backbone.

kion-dgl:
Closes #593. Closes #592 (item 1 was already done in `34dc9e94`).

## #593 — the walls

196 authored walls were imported and then thrown away by `PLACE_WALLS := false` inside `_to_object()`. That is **generation** time, and it was the wrong place twice over:

- the generated field **data** differed between an autopilot run and a player run, so the committed dump and `check_generated_objects.py` described a world no player sees;
- a cell first entered under autopilot **saved without its walls and restored that way forever**.

The skip now lives in `_spawn_wall()` under `PSZ_AUTOPILOT_NO_WALLS` — one guard covering both the fresh and restore paths, exactly like `PSZ_AUTOPILOT_NO_BOXES`.

**Its own variable, not folded into `NO_OBSTACLES`** (the issue asked me to pick and say which): `NO_OBSTACLES` strips a stage's *baked obstacle collision*, a property of the room mesh; this drops a *placed object*. A run that wants to prove walls don't wedge the backbone needs to turn off exactly one of them. Set in the three matrix scripts; `record_free_roam.sh` takes it from the environment and **defaults it OFF**, so the default run is the one that tests the walls.

### The soft-lock question is answered, and measured

The issue asked for one look at psz-re's `wall_placement_per_room.json` before implementing. It carries `IT_NEVER_BLOCKS_A_CONNECTION`: over all **964 records** the smallest distance from a wall to any doorway mouth of its own room is **5.14 cells** — with a control that the same ruler finds the room warp at **0.01** and `o0c_fence` at **1.31**, so it does detect things standing in a doorway.

`test_authored_walls_clear_doorways` re-derives that from **our** copies rather than trusting the number, sampling along each opening rather than just its endpoints:

```
wall           min-dist-to-doorway   5.75
gun_trap       min-dist-to-doorway   3.37   ← closest object of ANY kind
```

The import is ours to get wrong, so the invariant is ours to check.

**Destructibility is still open upstream** — psz-re: *"no wall behaviour is decoded — hit points, whether it breaks at all"* — so the spawner's `destructible` default is unchanged. The clearance above is what makes that choice low-risk either way.

## #592 item 2 — the unreachable layout index

Five layout masks, four reachable, **by design**, now recorded in both `the_rule` and `authored_objects()`'s doc comment so the next reader doesn't re-derive it: `FUN_02082814` returns a category 0..3 against the table at `0x020f2e70` (three rows of four, each summing to 100), validated on all 44 values across five captured fields. Mask `0x30` and its 146 group-4 objects belong to the mission-guarded path psz-re hasn't modelled.

## Not in the issues, but it had to happen today

**The importer now has an allowlist.** psz-re's object dump grew in-flight from 2,726 to **4,838** classified records — keys, fences, switches, mspack, NPCs, warps: the #594 work, currently uncommitted in that repo. Without a filter those kinds flow straight through as object types the spawner cannot build, and because they still count against the **20-object room cap**, an unhandled key would silently displace a box or a trap.

`CONTAINER_KINDS` + `TRAP_IDS` is now explicit; everything else is counted into a `not_imported_yet` block **in the file** so the gap is visible rather than silent (2,112 records sit there today). Widening it is #594's job, together with the spawner cases.

## Verification

- **Tests: 2563 passed, 0 failed.** The old `PLACE_WALLS` assertion was **inverted** rather than deleted — a generated cell must now *carry* its walls, since the gate moved out of the data. `check_generated_objects.py` likewise fails on their absence and reports **79 walls over 392 cells**.
- `code_graph --check`, `import_re_objects --check`, `check_repo_layout` — clean.
- Regression matrix with `NO_WALLS=1`: running at the time of writing; I'll comment with the result.

### Honest gap — this is #593's own acceptance criterion and it is not met

The free-roam autopilot passes with walls enabled (`DONE ok`) but **spawned zero authored walls**, because that harness loads the static `data/field_quests/*.json` — which carries only enemies, boxes and rare boxes — rather than `GridGenerator`. So that run is a **regression check, not evidence** that walls don't wedge the walk backbone.

The evidence that does exist is the clearance invariant above and the 79 walls in the generated dump. Getting the runtime half needs the free-roam harness pointed at the generator, which is its own piece of work — flagging rather than quietly claiming the criterion.

### DIFF
--- a/data/re_reference/room_objects.json
+++ b/data/re_reference/room_objects.json
@@ -9,7 +9,8 @@
   "the layout index is a rand(100) draw against a depth-banded weight row",
   "a mask is eligible only if every group it names is non-empty; group 5 is forced in",
   "group 5 is ROLLED: rand(100) against [40,20,20,20] for a count of 0..3, Fisher-Yates shuffle, take that many",
-  "a group is truncated to 20 records and a room stops at 20 objects"
+  "a group is truncated to 20 records and a room stops at 20 objects",
+  "THERE ARE FIVE MASKS AND ONLY FOUR ARE REACHABLE, BY DESIGN. The layout index comes from FUN_02082814, which psz-re reads as a weighted draw returning a category 0..3 against the table at 0x020f2e70 -- twelve ints as three rows of four, every row summing to 100 -- and validates on all 44 values across five captured fields being in 0..3. So mask 4 (0x30), the only one naming group 4, is never selected by the depth-banded draw and group 4's objects never spawn in a free field. That is faithful, not a bug: psz-re's NOT_ESTABLISHED records a mission-guarded path that forces layout 4, which it has not modelled and which free-roam generation correctly never takes. Do not 'fix' the mask table or widen the draw."
  ],
  "layout_masks": [
   33,
@@ -60,6 +61,35 @@
   "capture_trap": 66,
   "burn_trap": 43,
   "needler_trap": 16
+ },
+ "not_imported_yet": {
+  "_": "Kinds psz-re emits that this importer deliberately does not take yet -- psz-godot #594 widens CONTAINER_KINDS and adds the spawner cases together. Listed so the gap is visible rather than silent.",
+  "counts": {
+   "o0c_key": 836,
+   "o0c_trebox": 653,
+   "o0c_switchf": 107,
+   "o0c_mspack": 102,
+   "o0c_fence": 98,
+   "o0c_shfence": 41,
+   "o0c_needle": 41,
+   "o0c_remswitch": 26,
+   "o0c_dgfance": 26,
+   "o0c_switchs": 24,
+   "o0s_warpm": 21,
+   "o0c_fence4": 18,
+   "o0c_butterfly": 16,
+   "o0s_warps": 15,
+   "o0c_return": 15,
+   "o0c_healhp": 14,
+   "o0s_warpb": 14,
+   "o0c_warparena1": 14,
+   "o0c_dragonfly": 8,
+   "o0c_poisonm": 7,
+   "np_011_00_0": 5,
+   "np_014_00_0": 5,
+   "np_019_00_0": 5,
+   "o0c_bird": 1
+  }
  },
  "rooms": {
   "s00a_nr1_d": {
--- a/scripts/3d/field/cell_object_spawner.gd
+++ b/scripts/3d/field/cell_object_spawner.gd
@@ -1098,7 +1098,21 @@ func _spawn_bear_trap(pos: Vector3) -> void:
 	print("[CellObjects] BearTrap at %s" % pos)
 
 
+## Authored walls, skipped at SPAWN time under PSZ_AUTOPILOT_NO_WALLS.
+##
+## Deliberately its own variable rather than folded into NO_OBSTACLES: that one
+## strips a stage's baked obstacle collision, which is a property of the room
+## mesh, while this drops a placed object. A run that wants to prove walls do
+## not wedge the backbone needs to turn exactly one of them off.
+##
+## The skip is here and not in FieldPopulation because the gate must not change
+## the field DATA: a cell first entered under autopilot would otherwise save
+## without its walls and restore that way forever, and the committed
+## generated-fields dump would describe a world no player sees. Same reason and
+## same shape as PSZ_AUTOPILOT_NO_BOXES above.
 func _spawn_wall(pos: Vector3, rotation_deg: float, is_destructible: bool = true) -> void:
+	if OS.has_environment("PSZ_AUTOPILOT_NO_WALLS"):
+		return
 	var wall := WallScript.new()
 	wall.is_destructible = is_destructible
 	_c._map_root.add_child(wall)
--- a/scripts/3d/field/field_population.gd
+++ b/scripts/3d/field/field_population.gd
@@ -367,6 +367,20 @@ static func _to_object(rec: Dictionary) -> Dictionary:
 		snappedf(float(rec.get("z", 0.0)), 0.01),
 	]
 	var obj: Dictionary = {"type": kind, "position": pos, "authored": true}
+	if kind == "wall":
+		# OURS, and stated rather than inherited from the spawner's default.
+		# psz-re's wall_placement_per_room.json is explicit that "no wall
+		# BEHAVIOUR is decoded -- hit points, whether it breaks at all", so
+		# there is no measured answer to port. Breakable is the failure-safe
+		# reading: no authored wall blocks a connection today (5.14 cells of
+		# clearance across the corpus), but if one ever did, a player can clear
+		# a breakable wall and cannot clear a solid one.
+		#
+		# The thread to pull if this is ever settled: the container constructor
+		# special-cases record byte +0x12 == 1 -- it additionally enables the
+		# collider and stores 2 rather than 3 -- which is the likeliest
+		# solid/breakable discriminator. This importer does not carry +0x12.
+		obj["destructible"] = true
 	# The spawner's `rotation` is degrees everywhere (it calls deg_to_rad), so
 	# convert here rather than leaving a second unit in the object schema.
 	var facing_deg: float = float(int(rec.get("a", 0))) / FACING_PER_TURN * 360.0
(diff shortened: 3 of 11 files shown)


kion-dgl:
## Regression matrix: PASS (`rc=0`)

The result I owed this PR. Ran on `feature/authored-walls` with `PSZ_AUTOPILOT_NO_WALLS=1` set by the script, full chain (boot → freeroam → SR → PP → probes → AS → DOE → FO → SIS → IT → heretic → CSY → TBS → DC → shops).

Every gated phase reached `[sanity] DONE ok`:

```
dc          DONE ok      as_sol      DONE ok
tbs         DONE ok      as_moon     DONE ok
csy         DONE ok      as_canon    DONE ok
heretic     DONE ok      pp_backtrack DONE ok
it          DONE ok      pp_canon    DONE ok
sis         DONE ok      pp_dialogue_probe DONE ok
fo          DONE ok      freeroam gurhacia DONE ok
doe         DONE ok      freeroam rioh     DONE ok
```

**Two non-`DONE ok` entries, both pre-existing and both deliberately non-gating** — the script names them itself (`FREEROAM_GATED="gurhacia rioh"`, with ozette/paru "human-playable; door-geometry follow-up"):

```
freeroam_ozette  FAIL: walk stuck at dist=7.91 in cell 2,3 (s02b_lb1) — author waypoints for this stage
freeroam_paru    FAIL: walk stuck at dist=8.73 in cell 0,2 (s05a_lb1) — author waypoints for this stage
```

Checked rather than assumed to be unrelated: both are the documented "author waypoints" snag, and neither log mentions a wall (ozette 0 occurrences; paru's 1 is the unrelated `step-up on_wall` line). `pp_menu_probe` has no `DONE ok` by design — it is a probe with its own assertion.

Worth noting for the earlier caveat: those two failures are in **free-roam**, which is exactly the harness that loads the static `field_quests/*.json` and therefore spawns no authored walls at all. So the matrix confirms nothing regressed; it still is not evidence about walls on the walk backbone. That gap is unchanged.

kion-dgl:
Folded into #599 and closing this in its favour.

All three touched field generation, and this one plus the dead-doors fix both regenerate `web/public/field-dumps/generated-fields.json`. Merging them separately would have text-merged two independent regenerations into a dump neither generator would actually emit — and `check_generated_objects.py` would most likely still have passed on it. The only correct resolution is one regeneration from the combined code, which is what #599 carries.

Every commit from this branch is preserved in #599's history; nothing was squashed or rewritten. The review notes raised here are addressed there too.

Practical reason as well: one PR is one local Godot run to verify, instead of three.

Issue #595, opened 2026-08-13
Field gate economy: key gates, two-key gates, enemy-defeat gates and key placement (psz-re confirmed)


kion-dgl:
**Fully decoded upstream, zero blockers.** psz-re's `nodes/sys.field-doorways.json` is `confidence: confirmed`, validated over **62 field captures / 592 rooms / 1,058 connections**, and `data/level_topology_builder.json` carries the whole economy with real parameter values. This is the biggest gameplay change of the current field-generation batch, so per the working agreement it wants an Astro `/states` page as the normative contract **before** any `.gd` changes.

## The four door kinds

The door attribute lives at cell `+0x14+dir`, copied to room record `+0x1c..+0x1f`:

| Attr | Meaning |
|---|---|
| 0 | open |
| 1 | one-key gate |
| 2 | **two-key gate** |
| 4 | enemy-defeat gate |

## The rules, as measured

- **Key-gate budget**: `(roomCount - 2) * params[5] / 100`, `params[5] = 35`. Start and goal excluded. Hard cap — the pass stops as soon as spent reaches budget.
- **Eligibility** (`FUN_020b2978`): reject if the room already carries more than one gate, if **any neighbour carries a gate** (gated rooms are never adjacent), or if the room's gate/key balance is off. Rooms with flag `0x20` accept at `rand(100) < params[4]` (50); other rooms accept with a probability that **scales with remaining budget**, so gates thin out rather than stopping at a cliff.
- **One vs two keys** (`FUN_020b27ec`): roll `rand(100)` against `params[6] = 30` when the room's flag byte has `0x20`, else `params[7] = 10`. Under → attribute 2 (two keys), over → attribute 1. A two-key gate additionally requires the caller's count ≥ 2.
- **Enemy-defeat** (`FUN_020b2bf4`): a percent chance **per ROOM, not per door**. One roll; if it passes, *every* direction that has a neighbour, is not already attributed 7, and is not the way-back direction gets attribute 4. The way-back door is never gated — 0 of 592.
  ⚠️ `level_topology_builder.json` lists `default_params_8: 10`, but `level_generation_topology.json` records the **runtime** value as **75**, and the capture split (268 rooms gate every eligible exit, 106 gate none, 0 mixed ≈ 72%) backs 75. **Use 75.**
[...]

Issue #581, opened 2026-08-08
psz-re: do field rooms place walls, and is the room's own door set ever sealed?


kion-dgl:
**From psz-godot.** Two related geometry questions that came out of making free-field generation actually work.

## 1. Walls (`oNN_wall` / `oNN_wallptcl`)

Every field ships a destructible wall model and a matching break-apart debris model, re-skinned per area (`o01_wall` … `o07_wall`; the Eternal Tower has none). The game code has a `wall` object type with an `is_destructible` flag.

**No field places one today** — not the hand-authored quests, not the generated fields — and there is no RE data on wall placement at all.

- Do generated field rooms contain walls, or are walls quest/story-only furniture?
- If they are generated: what decides count and position, and does a wall ever block a *connection* (i.e. can it gate progress), or is it always cosmetic clutter within a room?

That last part is why we have not guessed: a wall blocks, so a wrong guess is a soft-lock rather than a cosmetic complaint. Traps only chip HP and we still got them visibly wrong.

## 2. Unconnected doors

Our generator now picks a room by **rotation** — it rotates a room model until its door set *covers* the connections the cell needs, which is what `scripts/tools/refield/derive_layout.py` already does for the observed layouts. Extra doors are left over, and we leave them as inert geometry: the runtime only builds a gate trigger for a direction that appears in the cell's `connections`, so the player cannot walk through one.

- **Is that what the original does?** Does PSZ pick rooms whose doors exactly match the required connections, or does it also allow extra doors?
- If extra doors are allowed, **does the game seal them visually** — a wall/shutter mesh over the opening — or does the door just stand there leading nowhere?

If it seals them, we would like to do the same; it is the difference between a room that reads as intentional and one that reads as a bug.

Related: #562, #578.

kion-dgl:
**From psz-re.** Both halves answered, and the second one comes back **refuted as posed** — with something better in its place.

Short version:

1. **Walls are placed, they are the box's own pipeline, and they never touch a doorway.** Build them; nothing can soft-lock.
2. **A room's door set is never sealed, because it never has a spare door.** PSZ picks the room by the cell's **degree**, so doors *are* the connections. Your rotation-fit rule produces a situation the original does not have.
3. **What the game does close is a GATE**, and always on a doorway that leads somewhere — the per-doorway byte psz-re already lists in §3 of the parity doc. That is the "shutter mesh" you were asking about; it just belongs on a *live* connection, never on a spare one.

[...]

PR #598, opened 2026-08-13
Spec: Field Gate Economy — the normative contract before any generator code (#595)

kion-dgl:
Step 1 of #595, and only step 1. The working agreement says the definition lands before the `.gd`, and that the tests the implementation owes get pinned against the page rather than against whatever gets written — so this is the contract, not the feature.

**Spec-only PR** (`spec/` + one comment correction), so no autopilot per the repo's own rule.

## What the page says

`/states/field-gates` — RFC-2119 — covering the four door attributes (`0` open, `1` one-key, `2` two-key, `4` enemy-defeat), the key-gate budget `(rooms - 2) * 35 / 100` as a hard cap, the eligibility rules that keep gated rooms from ever being adjacent, the one-vs-two key roll at `params[6] = 30` / `params[7] = 10`, BFS key scatter to depth < 2 capped at 2 per room, and the solvability invariant.

Two things the page leans on hard, because they are what an implementation gets wrong by default:

- **The enemy-defeat roll is per ROOM, not per door.** One roll; every eligible forward exit is gated or none is. The capture set: **268 rooms all-gated, 106 none, 0 mixed.** A per-door implementation would look similar in play and be wrong.
- **The chance is 75, not 10.** `level_topology_builder.json` lists `default_params_8: 10`, but `level_generation_topology.json` records the runtime value as 75 — and 268/106 is ≈72%, which is 75 with sampling noise and nothing like 10.

And the one that MUST NOT be skipped: the **solvability invariant** walks the way-back chain accumulating gate requirement vs keys available and requires them equal. It is simultaneously the most likely thing to drop as an optimisation and the most expensive to add back.

## The correction #595 forces

`grid_generator.gd:75` said extra doors beyond entry+exit are allowed and "a door with nothing behind it is inert geometry — which is already true of the static fields."

The first half still holds for *our* generator. The implication about the original does not. psz-re's `sys.field-doorways.json` exists to answer our #581 and finds **there is never a spare door**: shape comes from the cell's **degree** and nothing else, and the degree counter is bumped in the same function that links two neighbours — so **doors == connections by construction**, with `room_doorways.json`'s per-shape counts agreeing independently. Nothing is ever sealed; what the game closes is a **gate**, always on a live connection (417 of 417 still hold their neighbour index).

The comment now says the tolerance is a divergence we're *choosing* rather than parity, and points at the spec page.

## Reconciliation, not contradiction

`field-lifecycle.astro` already defines keys as **section-scoped and fungible** — any key opens any key gate in the section. The measured invariant is over **counts** (required vs available), not identities, so the two agree and the page says so explicitly.

The original *does* store per-room key ids, and what they're for is **not established**. The page forbids pairing keys to gates on the strength of them.

## Also stated as not-claimed

- the exact taper on the non-`0x20` acceptance probability (psz-re records only that it scales with remaining budget);
- what flag `0x20` means — it's used as an input, not explained.

Linked from the states index and flagged **"contract only — not yet implemented"** so nobody reads it as a description of the runtime. Astro builds clean.


kion-dgl:
Folded into #599 and closing this in its favour.

All three touched field generation, and this one plus the dead-doors fix both regenerate `web/public/field-dumps/generated-fields.json`. Merging them separately would have text-merged two independent regenerations into a dump neither generator would actually emit — and `check_generated_objects.py` would most likely still have passed on it. The only correct resolution is one regeneration from the combined code, which is what #599 carries.

Every commit from this branch is preserved in #599's history; nothing was squashed or rewritten. The review notes raised here are addressed there too.

Practical reason as well: one PR is one local Godot run to verify, instead of three.

PR #601, opened 2026-08-14
Place the authored fences, switches and the poison trap

kion-dgl:
Three of the four items from kion's list. **Enemy waves are deliberately not here** — see the end.

psz-re's per-room dump emitted 3 kinds out of 30 until `aef6d1b`; it now emits every classified kind, so the importer's allowlist can widen. The spawner has carried `fence` and `step_switch` cases for the static quests all along, so most of this is a mapping job rather than new gameplay code.

## What now appears in a generated field

| | set `d` records | placed in the dump |
|---|---|---|
| switches | 157 | 64 |
| fences | 115 | 54, across 20 cells |
| poison traps | 7 | 5 |

On top of the boxes, 138 walls and eight trap families already there.

## Fences needed a design call

The fence↔switch pairing is **not published**: psz-re's record carries `group/kind/model/x/y/z/angle/slot/factory_index/variant` and no link, because the linkage lives in the object parameter block that the per-room dump doesn't expose.

And a fence is not scenery. psz-re measures `o0c_fence` at **1.31 cells from a doorway**, against a 5.14 floor for walls — fences stand *in* doorways on purpose, because a fence is a barrier you open. So an unlinked fence is not a cosmetic bug, it is **a sealed room**.

Until the flag is published: every fence in a room is linked to every switch in it, and a fence with no switch is dropped rather than placed. Enforced at **two** layers, because one is provably not enough:

1. The **importer** refuses to emit a fence into a room whose table holds no switch — 33 rooms of set `d`, reported in `fences_skipped_no_switch` rather than silently dropped.
2. `_drop_unopenable_fences` re-checks the objects a room actually **builds**, because the layout mask can build the group holding the fence and skip the group holding the switch. Measured before the guard: **1 sealed cell** out of the 21 carrying a fence. After: **0**.

Replacing this with the real pairing touches those two functions and nothing else.

## The poison trap is a placeholder, as asked

`o0c_poisonm` is a real authored trap in the corpus with no behaviour of its own here yet. It rides the shared `needle_trap` script through `trap_kind`, like the other four contact traps — the object is in the world, and no poison mechanic is invented to justify it.

## Autopilot

Fences skip at spawn time under `PSZ_AUTOPILOT_NO_FENCES` — its own variable rather than folded into `NO_WALLS`, so a run can prove one without the other. Wired through all four autoplay scripts. Same shape as walls: the *data* is identical whether or not the autopilot is driving.

## Verification

| | |
|---|---|
| `test_runner` | **2575 passed, 0 failed** |
| new seeded test | no room roll leaves a fence unopenable, over 130 fence rolls |
| `check_generated_objects` | 392 cells, 801 containers, 343 traps, 138 walls |
| `import_re_objects --check` | matches psz-re at **3,005 objects** |
| `code_graph` / `orphan_files` | clean |
| `sanity-check` | running; will post the result |

## Not here: enemy waves

The cause is confirmed — `objects_for_cell` emits enemies with no `wave` key, so `_max_wave` is always 1 even though the runtime supports N waves. But no wave count is decoded: `enemy_room_assignment` gives each room its templates **with weights** (alternatives, not rounds), `enemy_wave_rolls`'s `still_open` says what drives the roll is unidentified, and the static RE-built `valley_field.json` has all 54 enemies on wave 1 too.

kion's play observation — **2–3 waves per room, 2–5 enemies each** — fits the data well, since set `d`'s templates are 4 ids each, so the natural reading is that a room draws the weighted list once per wave. That's now a specific question with psz-re, along with what makes enemies appear once the player is already in the room. Implementing it before that answer would be inventing content, so it waits.


kion-dgl:
`sanity-check`: **27/27 passed**, godot rc=0, recorded for `aaea177e`.

PR #618, opened 2026-08-16
Put load triggers in the doorway they belong to

### LINKED ISSUE #617: Load triggers still derive from the authored position, so they no longer line up with the moved gates
Follow-on to #612, which moved the gate mesh onto psz-re's measured doorway but deliberately left the spawn and load trigger deriving from the authored `position`.

That split was correct at the time and the reasoning is recorded in `portal_gate_manager.gd`:

> a trigger is 6 deep, so its near face is gate+4. A doorway stub only runs from the wall (|22|) to |25|, so moving `position` itself put the trigger's near face at |26| — past the stub, over nothing — and the autopilot walked to the edge and stopped (measured on s01b_tb3).

So navigation was protected. But the gate moved and the trigger did not, and the two are now measurably out of line.

## How far out

Measured across `data/stage_configs/unified-stage-configs.json` — 552 portals carry `gatePosition`, 50 do not:

| | offset between `gatePosition` and `position` |
|---|---|
| median | 1.88 |
| mean | 2.57 |
| p90 | 5.17 |
| p95 | 7.54 |
| max | **20.81** |
| portals over 4.0 | **100** |
| portals over 6.0 (a full trigger depth) | **43** |

Worst offenders:

```
20.81  s01b_xb2  (north)
16.92  s01a_ic3  (north)
16.27  s02a_sa1  (south)
15.11  s01b_td2  (south)
14.41  s01a_lc1  (east)
```

A trigger is 6 deep with its near face at gate+4, so **100 portals have a gate the trigger's near face does not even reach**, and on 43 of them the gate is a full trigger depth away. The player sees a doorway in one place and the cell load fires somewhere else — either before they reach the visible gate, or not when they walk through it.

## What the original gives us

`room_doorways.json` has the doorway geometry, and it is exact: each segment is the doorway axis, e.g. `s01a_ib1` → `[[-17.0, -22.0], [-17.0, -25.0]]`. The room wall is at `|22|` and the doorway stub runs out to `|25|` — **3 units deep**.

**It does not give us the original's trigger volumes.** There is no trigger or load-zone record in psz-re's data. So the honest target is "a trigger inside the original's doorway", not "the original's trigger" — the geometry is measured, the volume is ours.

## Proposed fix

Derive the trigger from `gatePosition` rather than `position`, and size it to fit the stub it now has to live in: **depth at most 3** (currently 6), centred around `|23.5|` — between the wall face and the stub end — instead of offset +7 from the authored position.

Spawn (+3 from `position`) is a separate question and should probably stay put until the trigger change is proven, so that only one thing moves at a time.

## Risk

This is the change that broke the autopilot before. `s01b_tb3` is the known failure case and is the first thing to run. A shallower trigger is easier to miss at speed, so the matrix — not a unit test — is what decides this. Waypoints may need re-authoring where the walk line assumed the old trigger position.

kion-dgl:
Closes #617.

## The bug

#612 moved the gate mesh onto psz-re's measured doorway and deliberately left the spawn and load trigger deriving from the authored `position`. That protected navigation at the time, but it **broke the line**.

A portal's three points are one line — **gate | spawn | trigger**, same lateral position, same axis, authored at `+0 / +3 / +7` outward. Moving only the gate left the other two behind by exactly the distance the gate travelled, on 552 portals.

## The fix

**Move the spawn and trigger by the same delta the gate moved.** Nothing is normalised, nothing is classified, and every authored relationship survives: `gate→spawn` stays 3.00 on every portal, including `s02a_sa1` whose gate moved 16.27 units.

`spawnPosition` and `triggerPosition` are written per portal, the same way `gatePosition` already was — the engine and `validate_graph.mjs` read them, so the geometry lives in data rather than being restated in GDScript, JS and Python.

## Three rules were tried. Two are recorded as refuted

| rule | how it failed |
|---|---|
| centre the trigger in the doorway stub | the box is 6 deep against a 3-deep stub, so 1.5 units protruded back into the room; the autopilot crossed it while still looting, cells loaded early, drops were abandoned, and `the_paru_pact` cleared with objectives unmet and no Telepipe |
| fixed distance from the gate (the median) | `s02a_tb3`'s gate moved 1.66 but its authored `gate→spawn` was 4.66, so snapping to the median dragged the spawn 3.56 units, shed an item drop, and `apothecary_supply` burned all three telepipe interacts picking it up |
| **move by the gate's delta** | **passes** |

Both wrong rules tried to derive the geometry from first principles. The authored offsets *were* the pattern; the fix is "change nothing except what #612 changed."

## Floors

Aligning the line pushed some triggers past the edge of their stage's floor, where nothing can reach them. Rather than find these one autopilot run at a time — three deep before it became clear the queue had no known length — `scripts/tools/refield/check_portal_floor_cover.py` answers it offline: it loads each floor mesh, projects to XZ, and point-in-triangle tests every spawn and trigger.

It tests the trigger's **near face**, not its centre: the box is 6 deep and the cell loads the moment the player crosses the face. Testing the centre reports 155 doorways, nearly all a uniform ~3 units past a ±25 floor — exactly the box's half-depth, not a gap. Against the face it was **5**, and the list was confirmed by two independent autopilot failures (`s01b_lb1` in DoE, `s03a_ga1` in FO) before either run hit them.

kion authored eight floors. **The checker is now clean: every spawn and trigger across all 552 portals in 296 stages lands on floor.**

A predictor worth recording as dead: "gate moved more than 8 units" flags 25 portals and is wrong about nearly all of them. `s01b_ic1`'s south moved **3.14** and broke; `s01b_lb1`'s east moved **1.90** and broke; `s05a_sa1` moved **13.4** and was fine. Distance is not the question.

## Also here

- **`QUEST_TIMEOUT` 900–1200s → 600s.** Nothing that passes has ever exceeded 270s, so a failing quest was burning 15–20 minutes per attempt and an hour across three. This was the single biggest drag on diagnosing #617.
- **The stuck-walk advice now matches the floor probes it already took.** `_log_floor_samples` prints `floor=false` on all three probes directly above a failure that then says "author waypoints for this stage" — which is exactly wrong when the answer is that there is nothing to walk on. It sent this PR chasing waypoints twice. `0/3` now reads "NO FLOOR ahead — extend the floor"; `1–2/3` reads "partial floor, a waypoint can route around it".

## Testing

**Full autopilot matrix, every phase passing, all first attempt:**

```
boot, freeroam gurhacia, SR          pass
PP canon / backtrack / dialogue      pass
AS canon / moon / sol                pass
DOE  319s   FO  257s   SIS  235s     pass
IT   129s   heretic 104s  CSY 108s   pass
TBS, DC                              pass
```

Before this branch, everything from PP onward failed and the endgame spine was skipped entirely.

- web **616 passed**, godot **2657 passed, 0 failed**, graph validator clean on **294 stages**
- new guards assert the line directly: fixed spacing from the gate, all three on one axis, and the spawn outside the trigger box — the last of which is the specific thing no earlier test could see, since "the box overlaps the doorway" stayed true through both refuted rules


kion-dgl:
## Autopilot found a real regression in this PR, and it is fixed

Running the matrix on the Linux box caught something no test here could see. Recording it because the bisect is the useful part.

### The failure

`the_paru_pact` cleared every room on an **identical 23-cell route** and then stranded: no Telepipe, no warp out, 14 minutes of timeout. Every downstream quest then failed on `guild accept limit reached` for want of a post-save. **One portal, twelve dead phases.**

### Isolating it

| run | code | data | pp_canon |
|---|---|---|---|
| control `00cb388b` | old | old | pass 216s |
| **isolation** | **this PR** | old | **pass 216s** |
| this PR | new | new | fail 3 × 1100s |
| this PR minus `s05b_ga1` | new | new−1 | fail |
| this PR minus `s05a_sa1` | new | new−1 | fail |
| **only `s05b_nc2` re-anchored** | new | +1 | **fail** |

The isolation run proved the engine change was innocent and that PP is deterministic, not flaky. Bisecting the data landed on **one portal out of 124** — `s05b_nc2`, which had been flagged as broken by **0.06 units**.

### The cause

The three portal points are authored **in order, outward: gate | spawn | trigger**. #612 moved the gate. This PR moved the trigger. **Neither moved the spawn.**

```
gate 22.0 | spawn 24.06 | trigger 28.06     authored, correct
gate 22.0 | trigger box 22.0 .. 28.0        after the first fix — spawn swallowed
gate 22.0 | spawn 23.00 | trigger 27.50     now
```

A player entering a cell appears at the spawn. Spawn them inside the load trigger and they re-fire a transition on arrival, so the room is **bypassed**: objectives unmet, the harness force-completes the quest, and the goal cell's Telepipe — authored with `spawn_condition: "quest_complete"` and **deferred until objectives complete naturally** — never fires.

The Telepipe was a downstream symptom, not a coupling to trigger geometry.

### The fix

Spawn and trigger are both measured from the gate now (`+1.0` and near face `+2.5`), so the order holds by construction and both sit inside the 3-deep stub. `spawnPosition` joins `gatePosition`/`triggerPosition` as data; the engine and `validate_graph.mjs` read it; the spawn **node** moves with it.

Two refuted attempts are recorded in the commits: centring the trigger in the stub (put 1.5 units of a 6-deep box back inside the room), and pushing the trigger past the spawn (backwards — it treats the spawn as fixed and left 64 portals unfixable).

### Verification

- **`pp_canon` pass 217s, first attempt** — against the control's 216s. `pp_backtrack` pass 222s.
- Free-roam gurhacia cleared to the boss; boot and SR pass.
- Graph clean on 294 stages; web 615; godot 2657 passed, 0 failed; fixer idempotent.
- New guard asserts `gate < spawn < trigger_near` directly — the thing no earlier test checked, since "the box overlaps the doorway" is still true of a box that swallows the spawn.

Full chain from PP through `dark_castle` is running now.

Issue #604, opened 2026-08-15
Objects and enemies placed over the void in generated fields


kion-dgl:
From play-testing the generated free-roam fields: **objects and enemies are being placed where there is no floor.** kion's report — "stuff tends to be placed in the void … there was a key hanging over the void, enemies will be standing on top of rocks, and waves of enemies will be split over voids."

Three separate causes, and they can be fixed independently.

## 1. Keys use a geometric midpoint

`portal_gate_manager._create_key_pickup` uses the cell's authored `key_position` when there is one, and otherwise falls back to **the average of the portal spawn points**:

```gdscript
# Fallback: midpoint between portal spawns
var sum := Vector3.ZERO
for p in portal_positions: sum += p
key_pos = sum / float(portal_positions.size())
key_pos.y = 0.5
```

Generated cells never carry `key_position`, so **every key in a generated field takes this path**. In an L-corner the midpoint cuts the corner; in a 3- or 4-door room it lands in the middle of whatever is there, floor or not. That is the key over the void.

## 2. Enemies are still ring-placed

`FieldPopulation.objects_for_cell` puts enemies on a ring of radius 5.0, and the comment still says the deploy table "has not been decoded to positions". **That is stale.** psz-re has `data/enemy_deploy_positions.json`: 13,417 occupied slots, room-local `(x, z)` plus elevation, same frame as `room_doorways.json`, with 11,402 of 11,511 spawns inside ±22 and a median radius of 9.7.

A blind ring has no idea where rock formations or gaps are, which is both "standing on top of rocks" and "waves split over voids". Using the authored positions should fix the placement and the grouping together.

## 3. Nothing validates against the floor

`check_generated_objects.py` asserts each authored position is inside the room box (44×44 plus 3-unit door stubs). It does **not** check that there is floor underneath. A position can be perfectly faithful to the table and still hang over a hole, because the room mesh has holes the table does not know about.

The stage `*-floor.glb` meshes are committed (whitelisted in `.gitignore` precisely so they ship with the binary), so a floor raycast — or an offline check against the floor mesh bounds — is possible without the asset pack.

## Suggested order

1. Enemy deploy positions (biggest visual win, data already exists).
2. Key placement — put keys on an authored/validated position rather than an average.
3. A floor check in `check_generated_objects.py` so this class of bug fails CI instead of a play-test.

PR #623, opened 2026-08-24
Free-field fixes: end-room spawns, gate persistence, breakable walls, multi-wave

kion-dgl:
Three bugs caught play-testing Valley (savestate dumper + live runs).

## 1. End rooms were empty
`FieldPopulation.objects_for_cell` suppressed enemies for `is_start` **or** `is_end`, but the goal is "not necessarily last", so the path's final cell is usually an ordinary combat room and it spawned nothing. Now only the start room is unconditionally empty; the real goal room carries no assignment and is empty on its own.

## 2. Generated gates must persist
An enemy-defeat gate that would start open still built a frame and only dropped the laser. The first attempt (hide the frame + skip building open gates) was **wrong**: "open" is read from the *live* enemy count, which goes false the instant a room clears, so cleared gates **vanished on re-entry**. Fixed: a gate whose room has a fight (per its raw generated objects — a persistent signal) is built on every visit and shown open once cleared, so it stays. Only a door whose room *never* had enemies (plus attr-0 and the way-back) renders no mesh. `field-gates` spec updated to the persistence rule.

## 3. Walls weren't breakable
Walls declared themselves destructible but had **no hurtbox**, so attacks never registered — the wall blocked the player forever. Added a hurtbox (as boxes have) and a **3-hit** counter.

## Verification
Live Valley run: previously-empty end rooms now fight; cleared gates persist and show open on re-entry; walls break on the third hit. Field-generation `.gd` changes → sanity/autopilot matrix applies before merge (Kion's Mac). Still open from the session: enemy slot elevation (#604 floaters) and keys dropping into the void.

### DIFF
--- a/scripts/3d/elements/box.gd
+++ b/scripts/3d/elements/box.gd
@@ -1,7 +1,8 @@
-extends GameElement
+extends DestructibleElement
 class_name Box
 ## Destructible container that can drop items when destroyed.
 ## States: intact, destroyed
+## Hurtbox + intact/destroyed state live in DestructibleElement.
 
 signal destroyed_box
 
@@ -14,12 +15,6 @@ signal destroyed_box
 
 ## Drop amount for meseta or item ID for items
 @export var drop_value: String = ""
-
-## Collision body for physical presence
-var collision_body: StaticBody3D
-
-## Hurtbox for receiving hits from player attack hitbox
-var hurtbox: Hurtbox
 
 
 var _reticle: Node3D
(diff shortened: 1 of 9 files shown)


kion-dgl:
Added multi-wave rooms to this branch so everything's testable together (testing > separation for now). The wave runtime already existed (spawner reads a per-enemy `wave` field, controller spawns the next wave on clear); generation just never produced wave > 1. `objects_for_cell` now rolls a wave count and tags each wave. Count is provisional `[1,3]` until the real per-room `w3/w4` is imported from psz-re (.36 was down) — the code reads a `wave_counts` map when present. Spec for it is #622.

Issue #642, opened 2026-09-07
[FEATURE] Tower ga1: strip the phantom second portal, floors as 1-door terminals with in-room warps


kion-dgl:
## Description

Split out of #641 (scope decision there: tower is a separate design pass). The same phantom-portal correction and goal-terminal rule apply to the six tower floor records, plus a tower-generator rework:

1. **Data**: `s081`–`s086_ga1` each carry the phantom second portal — the bare 4-field `north` entry (no gate/spawn/trigger positions) whose coordinates coincide with the original's authored in-room warp object (`o0s_warpm` in `data/re_reference/room_reference.json`), not a doorway. Same transcription story as the field records fixed in #641.
2. **Generator**: `generate_tower_field` (`scripts/3d/field/grid_generator.gd`) keeps `ga1` in `TOWER_ROOM_TYPES` as an ordinary floor room with a doorway warp (`south`/`west`). In the original, a tower floor is a single 1-door room: you enter through the one measured doorway and leave by the in-room warp.
3. **Quest**: `investigate_tower.json` routes live connections through the phantom (e.g. the s081_ga1 start cell connects north via the phantom north portal); after the config fix these cells break unless re-authored. `generate_field_quest.py` also has a `boss_stage = "s086_ga1"` fallback to revisit.

## Evidence (all in `data/re_reference/`)

- `room_doorways.json`: all six tower `ga1` floors measure **exactly one doorway** (part of the 20-of-20 ga1 measurement cited in #641).
- `room_reference.json`: each tower ga1 room carries an authored in-room warp object — the phantom portals sit on those positions.
- Field-side precedent: #641 landed the same correction for the 14 field ga1 records (portal, waypoints, waypointEdges, `defaultSpawn` at the warp-object position) and reserved ga1 as the a/b section terminal; the free-roam smoke proves the in-room pad path end-to-end.

## Proposed implementation

- Config: strip the phantom portal + paired north waypoints/edges from `s081`–`s086_ga1`; add `defaultSpawn` at each room's measured warp-object position (pad anchor).
- Tower generator: ga1 floors become 1-door terminals whose floor-to-floor transition is the in-room goal pad (`AreaWarp_goal_pad` machinery); decide how non-ga1 floor rooms (`sa1`/`ib1`/`lb1` in `TOWER_ROOM_TYPES`) chain — they keep doorway warps today.
- Re-author `investigate_tower.json` + its quest plan (ga1 cells as terminals; no connection through the phantom).
[...]

Issue #639, opened 2026-09-06
[FEATURE] Key rework for exploration: drop on room clear, prefer dead ends


kion-dgl:
## Description

Two coupled changes to the key economy, both aimed at making exploration the reward loop:

1. **Keys drop when a room is cleared** (last enemy defeated), instead of sitting on the ground at their authored slot when the room loads.
2. **Key scatter prefers dead-end rooms**, so a dead end reads as a chosen side objective rather than forced backtracking.

## Motivation

Playtest design rationale (kion, after the backtracking fix landed):

- The original's per-room treasure boxes were already removed deliberately: walk back to the box, hit it, pick up loot, continue — it slows gameplay down and reads as a participation trophy when the focus should be on exploration.
- The original (and our faithful port, `_key_scatter_pool`) puts keys on the ground, so a dead end is just a wall you walk back from — "forced backtracking once you find it's not the way to continue."
- Dropping keys from combat deletes a rote pickup interaction and lets the fight deliver the reward; placing keys in dead ends converts dead ends into purposeful detours. Together: dead end visible → fight → key drops → gate ahead opens. That's an exploration reward structure, not a participation trophy.

Both are **intentional divergences from the RE-measured original rule** and must be recorded as such in the spec (same category as the treasure-box removal).

## Proposed Implementation

Per the repo's spec-first flow — spec + solver first, then runtime:

**Phase 1 — spec:** amend `/states/field-gates` (scatter rule gains a dead-end weighting, documented as a deliberate divergence from the measured BFS-depth<2 uniform rule) and `/mechanics/key-placement` (spawn condition becomes room-clear rather than room-load; authored slots still decide *where* a key appears). Update the captured-example text in `/states/free-field` to match.

**Phase 2 — solver (generator):** `_key_scatter_pool` / `_scatter_keys` in `scripts/3d/field/grid_generator.gd` weight the pool toward dead-end cells (degree-1 rooms that aren't start/goal). Start with dead ends *within* the existing BFS depth<2 pool (small divergence, tight economy); extending to any pre-gate dead end is a follow-up knob if detours feel too shallow in play.

**Phase 3 — runtime:** `_create_key_pickup` spawns on room clear instead of room load (drop at the clear event; authored slot stays the position). Constraint: **a key room that never rolls enemies must not soft-lock its gate** — runtime fallback (no enemies ever spawned in a key cell → key spawns on entry as today) plus a generator-side invariant/test that key cells are never empty.

[...]

Issue #422, opened 2026-06-28
Minimap: track live enemies as moving orange dots


kion-dgl:
**Source:** [playtest report 2026-06-28T05-55-10](https://psz.onl/reports/2026-06-28T05-55-10-095Z-247cfcef) — Rozalin, v0.37.15

## Observed
> "Just like how the player is tracked on the top right minimap as an arrow, enemies on the map should be tracked as moving orange dots until they are killed."

## Proposed spec (enhancement)
The field minimap MUST render live enemy markers:
- Each **alive** enemy in the loaded cell MUST appear as a marker (orange dot) at its minimap-projected position.
- Markers MUST update as enemies move and MUST be removed when the enemy dies.
- Marker style MUST be visually distinct from the player arrow (filled orange dot vs. directional arrow).
- **Bosses** MUST use a larger dot — **≈8× the area of a normal-enemy dot** (tester Sazonde99).
- Enemy markers MUST mirror the **existing player-marker scoping exactly**: scoped to the currently loaded cell, **no marker cap** (count is simply "however many enemies are alive in the cell"), and **no markers for off-cell enemies** (they aren't present).

## Likely code
`scripts/3d/field/room_minimap.gd` and/or `scripts/3d/field/grid_minimap.gd` already project the player to minimap space — reuse that projection for enemy transforms. Enemy roster is available via `scripts/autoloads/enemy_spawner.gd`. Add an enemy-marker layer that subscribes to spawn/death so dots appear/clear with the live roster.

## Tests (two-layer)
- **Unit:** spawn 3 enemies in a seeded cell, assert the minimap reports 3 enemy markers; kill one, assert 2. Spawn a boss, assert its marker scale ≈8× a normal dot.
- **Autopilot probe:** in a combat sanity run, log enemy-marker count and assert it tracks the live enemy count.

---

## Expected behavior (RFC-2119 spec contract)

| Enemy event | Minimap marker behavior (MUST) |
|---|---|
| Normal enemy present in loaded cell | a filled **orange dot** appears at its minimap-projected position |
| **Boss** present in loaded cell | a filled orange dot **≈8× the area** of a normal-enemy dot |
| Enemy moves | the marker position updates to follow it |
| Enemy dies | the marker is removed |
| Enemy not in the currently loaded cell | **no marker** (scoped exactly like the player/room today) |

- Enemy markers **MUST** be visually distinct from the player marker (filled orange **dot** vs. directional **arrow**).
[...]

PR #643, opened 2026-09-07
Goal rooms: reserve ga1 as the section terminal, warp in-room, fix the phantom second door

### LINKED ISSUE #641: [FEATURE] Goal rooms: reserve ga1 as the section terminal, warp in-room, fix the phantom second door
## Description

Three coupled corrections, all aimed at the section exit reading as a destination rather than a door bolted onto whatever room rolled last:

1. **Data**: every `a`-variant `ga1` stage record carries a phantom second portal; the original measures exactly one doorway.
2. **Generator**: reserve the set's `ga1` as the once-per-section terminal (a 1-door leaf), never an ordinary mid-path room.
3. **Runtime**: the section-exit warp dresses as the goal room's own in-room warp (the `AreaWarp_goal_pad` path already exists) instead of a doorway `AreaWarp` on a spare portal.

## Motivation

Playtest observation (kion): depending on the room, the area warp to the next section can look out of place. It currently is: the generator ends a/b sections on whatever room the path finished in (`lb3`, `ib2`, `ib1` in the current dump) and hangs the warp on that room's unused doorway. The original has a dedicated room type for this — the goal area — so the warp always lives in a room built to hold it.

## The rule, from psz-re

Four independent confirmations, all already in `data/re_reference/`:

- **Naming tables** (`level_room_naming.json`, arm9 `0x020F24AD`): the shape letter is ordered by connectivity — `i`/`l` 2-door, `t` 3-door, `x` 4-door, then `n`/`s`/`g` the 1-door terminals: dead end / **s**tart / **g**oal. `ga1` parses as "goal room".
- **Doorway measurements** (`room_doorways.json`, from the ROM's `lndpr` tables): **20 of 20** `ga1` rooms — s01–s07 a+b and the s08x tower floors — measure exactly **one doorway** (a north-wall opening; s01a at offset −6, s01b at +5).
- **Captured layouts** (`level_layouts_observed.json`, four savestates): exactly one goal room per section, and in `level_graph_observed.json` it is always a graph **leaf** — valley/wetlands/snowfield behind enemy-defeat gates, paru behind a two-key gate. One nuance: the goal room is not always the *list-last* entry (valley and snowfield list a dead-end after it) — the table order is generation order, not depth; the rule is "the section's terminal is the ga room".
- **Static quests corroborate by area**: valley, wetlands, snowfield and paru — the four captured areas — end their a/b sections in `ga1`. The three uncaptured areas (ruins, shrine, arca) end in ordinary rooms (`lb1`/`lc1`/`ic1`), authored before the layout data existed.

## Where we diverge today

- The generator's end cell is the last path cell of whatever shape rolled; the warp takes a spare doorway of that room.
- `ga1` sits in the ordinary room pool and **spawns mid-path** as a 2-door corridor room (current dump: `s01a_ga1` at a mid-path cell with two connections).
- Enabling bug: `s0Xa_ga1` carries portals `north`+`south` in `data/stage_configs/unified-stage-configs.json`, in the generator's `GATES` fallback (note `s01a_ga1: ["north","south"]` sits right next to a correct `s01b_ga1: ["south"]`), and in the authored quest cells it propagated into. All `b`-variants are correct. Likely transcription story: the config's "south" portal for `s01a_ga1` sits at x≈−6.13 — the same −6 offset as the original's measured *north*-wall doorway — so the real opening was probably placed but mislabeled, then a second phantom portal added to give the room an in+out.

## Proposed Implementation

Per the spec-first flow:

**Phase 1 — spec:** record the terminal rule on `/states/free-field` (one goal room per a/b section, a 1-door leaf, the section exit lives inside it) with the psz-re citations above; note the goal room MAY sit behind a key gate (paru measured one behind a two-key gate).

**Phase 2 — data:** correct the a-variant `ga1` records to the measured single door — `unified-stage-configs.json`, the `GATES` fallback, and the static quest cells where the phantom portal propagated (valley/wetlands/snowfield/paru sec-0/2 end cells). Verify first whether the imported stage assets themselves model a second doorway opening (floor stub / mesh gap) or only the portal tables claim it — if the asset diverges too, the geometry needs the same correction. Watch the mislabeled-portal lead: the "south" portal at the measured offset may just need relabeling north.

**Phase 3 — generator:** the end cell of every a/b section takes the set's `ga1` (degree-1 leaf; `e`/`z` keep `ia1`/`na1` as today). `ga1` leaves the ordinary pool. The retile already guarantees doors == connections, so a 1-door terminal is consistent with the no-spare-doors rule.

**Phase 4 — runtime:** the section-end warp renders as the goal room's in-room warp (the `AreaWarp_goal_pad` machinery the static paru quest already uses) rather than a doorway `AreaWarp`.

## Testing

- Seeded unit tests: every generated a/b section's terminal is its set's `ga1`, exactly once, never mid-path; `ga1` door count == 1 across configs and the `GATES` fallback.
- The generated-field free-roam autopilot smoke (folded into #640) walks each area end-to-end and exercises the in-room warp exit per section — it should stay green unchanged through this.

## Dependencies

- After #640 lands (the generated-field smoke is the end-to-end validator).

## Acceptance Criteria

- [ ] Every `ga1` record carries the measured single (north) door — configs, `GATES` fallback, static quest cells
- [ ] Every generated a/b section ends in its set's `ga1`, exactly once per section; `ga1` never mid-path
- [ ] The section exit renders inside the goal room, not on a doorway of an arbitrary last room
- [ ] Spec page records the terminal rule with the psz-re citations
- [ ] Generated-field autopilot smoke green per area

kion-dgl:
Closes #641. Tower floors split to #642.

> **Stack note:** this branch is based on the local `feature/key-rework-exploration` line (the unmerged #639/#640 commits plus the field-solver web tool — 7 commits before this PR's 3). They are dependencies (#640's smoke is the end-to-end validator), so they ride along here; merge order or a rebase is fine.

## What changed

**Data** — every a-variant `ga1` stage record (plus `s06b_ga1`, which the issue's "all b-variants are correct" missed) carried a phantom second `north` portal. Its coordinates coincide with the original's authored **in-room warp object** (`o0s_warpm`/`o0s_warpb`, `room_reference.json`) — a warp pad mistaken for a doorway, then given a portal entry. Stripped the portal, its paired north waypoints and `waypointEdges`; added `defaultSpawn` at the measured warp-object position so the section exit stands where the original puts it. `GATES` fallback and the two stale `stage_specs` follow.

**Authored fields & quests** — the four captured areas' a-end goal cells drop the phantom ref (their warp becomes the in-room pad, matching their b-ends; paru's a-end flips 180° so entry uses the real doorway). The mid-path ga1 usages authored before the layout data existed (ruins/shrine/arca fields, eight guild quests) swap to the area's i-corridors with real portals; `quest_plans` synced cell- and step-wise. Zero phantom portal references remain anywhere in quest data.

**Generator** — `ga1` leaves the ordinary pool (mid-path rolls, branch dead ends, retile candidates) and becomes the reserved a/b terminal: a 1-door leaf whose only door is the entry connection. The warp edge stays the nominal forward direction — a direction the room's portals do not carry — which is exactly what makes the field controller build the in-room goal pad instead of a doorway warp. e/z keep `ia1`/`na1`; the 5-room fallback now ends both areas on ga1.

**Spec** — `/states/free-field` records the terminal rule with the four psz-re confirmations (naming table, 20-of-20 doorway measurements, exactly-one-goal layouts, always-a-leaf graphs + the key-gate nuance) and the phantom-portal correction note.

## Verification

- Godot suite **3563 passed / 0 failed**, incl. the new `test_generated_goal_room_terminal` seeded sweep (7 areas × 3 rolls: exactly one ga1 per a/b section, never mid-path, portal-less warp edge, 1-door leaf) and re-pinned wetlands/GATES assertions.
- Web vitest **666/666** after the waypoint-edge cleanup; field dump regenerated — all 56 a/b terminals are their set's ga1, zero mid-path.
- Free-roam autopilot smoke (headless): **seed 6 clears gurhacia end-to-end** — a-section `s01a_ga1` (one open door) → `walk to goal pad at (2.0, 1.0, -15.3)` → e-transition → b-section `s01b_ga1` → goal pad → boss → `DONE ok`. The pads sit at the authored warp coordinates.
- The smoke's per-seed walk tail (floor ends / waypoint-to-author wedges in ordinary ib/lc/lb rooms) reproduces **identically on the base commit** (base fails seeds 2–3 with the same class; the RNG stream shift makes each branch roll different layouts per seed) — pre-existing and unrelated to ga1, which never appears in a wedge. Same for the search_and_rescue headless quest walk: byte-identical failure on base, upstream of every re-authored cell; the synced plans parse and drive up to that point. The dev-box recorded matrix remains the deeper quest validator, as before.

## Notably NOT changed

- The goal pad stays **free-roam-only** (quest guard untouched) — quests keep doorway warps per the issue discussion.
- Tower `s081`–`s086_ga1` keep their records and generator path → #642.
- ruins/shrine/arca section terminals remain the ordinary rooms they were authored with (only their illegal mid-path ga1 usages were swapped); re-terminating them is a separate authoring pass.


kion-dgl:
Playtest follow-up: key pickups now draw as gold diamonds on the room minimap (they drop on room clear since #639, but nothing marked where they stood). Marker appears with the drop, vanishes on pickup — same live-tracking pattern as the #422 enemy dots. Suite 3567/0 with the new `test_minimap_key_markers`.

kion-dgl:
Playtest fix #2: the goal pad floated a metre above the floor — its anchor is the `defaultSpawn` portal entry, whose parse paths hardcode y=1.0 for warping-in *players*, while the warp model's origin is at its base. `_spawn_goal_pad_warp` now grounds the node/trigger/mark at y=0. Suite 3567/0.

